PR #25, opened 2025-07-11
Fix enable_eplb missing arguments and add fp8 llama4 and qwen3 to CI

xuechendi:


xuechendi:
Failing is introduced by very recent upstream PR, submitted fix as below
https://github.com/vllm-project/vllm/pull/20849

PR #32, opened 2025-07-16
Proper chunked prefill bucketing/warmup

kzawora-intel:
Clone of https://github.com/HabanaAI/vllm-hpu-extension/pull/295


kzawora-intel:
stale

PR #61, opened 2025-08-06
Merged prefill fix - pad logits for sampler

adobrzyn:


xuechendi:
@adobrzyn , please help to elaborate the issue and how fixing works here, thanks

PR #62, opened 2025-08-06
Add t.compile config

Kacper-Pietkun:


xuechendi (commented):
@mswiniarsk , may you help to review this one

Kacper-Pietkun:
> @Kacper-Pietkun , May you help to add simple UT for t.compile as well
Sure, I added a test which covers regional compilation

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -2346,8 +2341,7 @@ def warmup_model(self) -> None:
 
         if not htorch.utils.internal.is_lazy(
         ) and not self.model_config.enforce_eager:
-            multiplier = 3 if os.getenv('VLLM_REGIONAL_COMPILATION',
-                                        'true').lower() in ('1', 'true') else 1
+            multiplier = 3 if self.compile_config.regional_compilation else 1
```
mswiniarsk, line 2344: I see that on vllm-fork the multiplier is being changed to `5`. Shouldn't we change it here as well? 

Also, why `3` or `5` do you know? Shouldn't it be somehow deduced from config what should be cache limit?

PR #65, opened 2025-08-07
Add INC dynamic quant test

yiliu30:
Add INC dynamic quant test for deepseek-v2.

cc @hshen14 @thuang6


hshen14:
pls do validate the R1 locally, as R1 might be too heavy for CI.

xuechendi (approved):

PR #68, opened 2025-08-08
Enabled structured output

tianmu-li:
1. Combine logits from decode and prompt
2. Move logits to cpu
3. Apply bitmask on cpu logits
4. Move logits back to hpu
All of these are only triggered when doing structured output


tianmu-li:
Accuracy from https://github.com/vllm-project/vllm/tree/main/benchmarks#%EF%B8%8F-example---structured-output-benchmark aligned with CUDA.

tianmu-li:
@xuechendi Fixed issue with merged prefill and rebased

xuechendi (approved):

xuechendi:
jenkins is failing due to known issue, merge this pr

PR #70, opened 2025-08-08
Fixed Plugin Test

slokesha:
Added main() to avoid python multiprocessing runtime error


michalkuligowski (approved):

slokesha:
Without main error:
image
With fix:
image
[plugin_log.txt](https://github.com/user-attachments/files/24060777/plugin_log.txt)

PR #75, opened 2025-08-12
Fix the problem that upstream tests don't release Gaudi cards after run

mandy-li:
As title. Delete llm instance at the end of run to release Gaudi cards.


xuechendi:
And the upstream test is triggered in docker, so I think we can leave the script as it is?

mandy-li:
> And the upstream test is triggered in docker, so I think we can leave the script as it is?

Even if it runs in docker, the cards won't be released if you don't kill the docker. If we want to continue to run other workloads in the same docker, we have to manually kill the processes.

xuechendi:
> > And the upstream test is triggered in docker, so I think we can leave the script as it is?
> 
> Even if it runs in docker, the cards won't be released if you don't kill the docker. If we want to continue to run other workloads in the same docker, we have to manually kill the processes.

we do remove the docker in upstream test => Current script is working OK, and will not block upstream PR even when some issue occurs and gaudi card gets seg-fault and can't get released.

=> So I suggest we don't change the current way.

I think you mentioned you was using this as example, might add another example in examples folder instead of using this one?

PR #90, opened 2025-08-19
Reduce number of compilations when dynamic shapes is used

anko-intel:
It fixes the issue with to many compilation for Pytorch dynamic shapes ( when VLLM_T_COMPILE_DYNAMIC_SHAPES=1)
It allows making dynamic shapes for registered buffers  (see UnspecializedParamBufferSource in PyTorch) by setting dynamo config.

### DIFF
--- a/vllm_gaudi/ops/hpu_rotary_embedding.py
+++ b/vllm_gaudi/ops/hpu_rotary_embedding.py
@@ -40,6 +40,11 @@ class HPURotaryEmbedding(RotaryEmbedding):
                                           2,
                                           dim=-1,
                                           output_size=cos_sin.shape[-1])
+
+        if not torch.compiler.is_compiling():
+            torch._dynamo.mark_dynamic(cos, 0)
+            torch._dynamo.mark_dynamic(sin, 0)
+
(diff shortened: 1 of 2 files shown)


vllm_gaudi/ops/hpu_rotary_embedding.py
```diff
@@ -40,6 +40,11 @@ def prepare_cos_sin(self,
                                           2,
                                           dim=-1,
                                           output_size=cos_sin.shape[-1])
+
+        if not torch.compiler.is_compiling():
```
Kacper-Pietkun, line 44: I think that this "if statement" is unnecessary as `prepare_cos_sin` method is basically always compiled (It is a part of `LlamaDecoderLayer` and its "additional" usage is compiled via `_rotary_prepare_cos_sin` in `HpuModelAdapter`. 
One exception would be if user for some reason decides to turn off regional compilaion. In that case `_rotary_prepare_cos_sin` will not be compiled, so the first usage of `prepare_cos_sin` also won't be compiled (to avoid this, we could call `_compile_methods` even when regional compilation is turned off).

PR #95, opened 2025-08-22
Enable sampler compilation

Kacper-Pietkun:


Kacper-Pietkun:
It can be merged only after the sampler warmup phase is created and merged.

PR #97, opened 2025-08-22
Warmup fix - for non contiguous PA runs, don't take more context blocks than possible

adobrzyn:


vllm_gaudi/extension/bucketing/linear.py
```diff
@@ -190,23 +190,29 @@ def generate_prompt_buckets(bs_bucket_config,
...
        and use_contiguous_pa:
         blocks_bucket_config[2] = max_blocks
     block_buckets = warmup_range(blocks_bucket_config)
-    if os.environ.get('VLLM_DECODE_BLOCK_BUCKET_MAX') is None\
+    if user_max is None\
        and max_blocks not in block_buckets and use_contiguous_pa:
         block_buckets.append(max_blocks)
     last_bucket = max_blocks
     for bs in bs_buckets:
+        max_blocks_including_max_model_len = bs * max_model_len / block_size
```
madamczyk-intel, line 207: float division. I guess it would be better to round it up:
```
max_blocks_including_max_model_len = bs * math.ceil(max_model_len / block_size)
```

adobrzyn: You're right

vllm_gaudi/extension/bucketing/linear.py
```diff
@@ -190,23 +190,29 @@ def generate_prompt_buckets(bs_bucket_config,
 
 
 def generate_decode_buckets(bs_bucket_config, blocks_bucket_config,
-                            max_blocks):
+                            max_blocks, max_model_len, block_size):
     buckets = []
     bs_buckets = warmup_range(bs_bucket_config)
     use_contiguous_pa = get_config().use_contiguous_pa
+    user_max = get_config().VLLM_DECODE_BLOCK_BUCKET_MAX
     if os.environ.get('VLLM_DECODE_BLOCK_BUCKET_MAX') is None\
```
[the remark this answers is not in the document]
adobrzyn: Removed that, not needed

PR #100, opened 2025-08-22
initial port

hsubramony:
port nixl

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,5 +65,6 @@ setup(
     entry_points={
         "vllm.platform_plugins": ["hpu = vllm_gaudi:register"],
         "vllm.general_plugins": ["hpu_custom_ops = vllm_gaudi:register_ops"],
+        "vllm.general_plugins": ["hpu_custom_connectors = vllm_gaudi:register_connector"],
     },
 )
--- a/tests/v1/kv_connector/nixl_integration/requirements.txt
+++ b/tests/v1/kv_connector/nixl_integration/requirements.txt
@@ -1,0 +1,4 @@
+pytest
+nixl
+lm-eval
+lm-eval[api]
--- a/tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
+++ b/tests/v1/kv_connector/nixl_integration/run_accuracy_test.sh
@@ -1,0 +1,222 @@
+#!/bin/bash
+set -xe
+
+# Models to run
+MODELS=(
+    "Qwen/Qwen3-0.6B"
+)
+#MODELS=(
+#	"meta-llama/Llama-3.1-8B"
+#)
+
+export VLLM_USE_V1=1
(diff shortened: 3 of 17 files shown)


hsubramony:
the latest commit works with vllm commit 17373dcd93ca60554d72cef4e159e70abbfd15af

vllm_gaudi/v1/worker/hpu_worker.py
```diff
@@ -136,7 +137,7 @@ def stop_profile(self):
 
     def init_device(self):
         # Initialize the distributed environment.
-        init_worker_distributed_environment(self.parallel_config, self.rank,
+        init_worker_distributed_environment(self.vllm_config, self.rank,
```
xuechendi, line 140: update to align with vllm latest

requirements.txt
```diff
@@ -6,4 +6,5 @@ tabulate
 setuptools>=77.0.3,<80.0.0
 setuptools-scm>=8
 numba
-transformers>=4.1,<4.56.0
\ No newline at end of file
+transformers>=4.1,<4.56.0
+nixl==0.4.1
```
yeonsily, line 10: @xuechendi The latest is 0.5.1 and it's not working for us and we use 0.5.0. Do you think we can update this to 0.5.0?

xuechendi: Does 0.5.0 works for you? Will that cause the NIXL_BACKEND_ERROR?

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -1159,27 +1182,47 @@ def _get_prompts_and_decodes(
...
+                not self.is_decoder_only(req_id):
                 # This is prompt
                 break
 
             # This is decode
             # NOTE(chendi): To support spec decode,
             # we don't assume num_scheduled_tokens == 1.
             # assert num_scheduled_tokens == 1
+            if not self.is_decoder_only(req_id):
+                assert num_scheduled_tokens == 1
```
xuechendi, line 1225: This will break assumption for spec decode, please not assert

xuechendi:
@hsubramony , please fix for failed CI and pre_commit

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -1941,6 +1985,8 @@ def _prepare_inputs(
             num_scheduled_tokens.append(seq_num_scheduled_tokens)
             num_prompt_tokens.append(seq_num_prompt_tokens)
             # NOTE: assert that all the decodes are "decodes".
+            if idx < num_decodes and not self.is_decoder_only(req_id):
+                assert seq_num_scheduled_tokens == 1
```
xuechendi, line 1989: Don't assert, which will fail spec decode

PR #102, opened 2025-08-25
[Upstream fix] Fix after #22711

adobrzyn:


vllm_gaudi/v1/worker/hpu_model_runner.py
mswiniarsk, line 1: You need to update also _execute_dummy_scenario: 
https://github.com/vllm-project/vllm-gaudi/blob/main/vllm_gaudi/v1/worker/hpu_model_runner.py#L2528
https://github.com/vllm-project/vllm-gaudi/blob/main/vllm_gaudi/v1/worker/hpu_model_runner.py#L2541


And also vllm-gaudi/tests/unit_tests/worker/test_hpu_model_runner.py:
https://github.com/vllm-project/vllm-gaudi/blob/main/tests/unit_tests/worker/test_hpu_model_runner.py#L150

mswiniarsk (changes requested):

PR #121, opened 2025-09-02
Do not add max_blocks as a bucket in linear bucketing with cpa

mswiniarsk:
The bucket is not needed and adding it ends with an error: 
`Worker failed with error 'could not broadcast input array from shape (4607,) into shape (2304,)'`


adobrzyn (approved):

adobrzyn:
Resolved with: https://github.com/vllm-project/vllm-gaudi/pull/126. I think we can close it

PR #131, opened 2025-09-03
Introducing sampler warmup as separate warmup step

ksmusz:
Warming up the sampler with different configurations removes graph recompilations of bigger sampler graphs seen within the actual execution. As tested with example workloads and batch sizes, the only recompilations left from the sampler are from minor graphs, which have minimal influence to the execution time.

The warmup of the sampler takes around 1-3 seconds, depending on the buckets' batch sizes to be warmed up.

Additionally, removed the situation, where the warmup method is called twice (seen as duplicated prints within the warmup phase but with empty warmed up buckets, as these have all been already warmed up).


sys-hab-pt-service:
Only codeowners can request to run Gaudi tests. Contact list: kzawora-intel, xuechendi, mswiniarsk, adobrzyn

PR #133, opened 2025-09-03
Unified Attention POC

madamczyk-intel:


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -1940,19 +2060,23 @@ def _prepare_inputs(
...
+            return (self._prepare_unified_prefill_inputs(
+                num_prefills, num_decodes, num_scheduled_tokens),
+                    self._prepare_unified_decode_inputs(
+                        num_decodes, num_scheduled_tokens))
+        else:
+            return (self._prepare_prefill_inputs(num_prefills, num_decodes,
+                                                 num_scheduled_tokens),
+                    self._prepare_decode_inputs(num_decodes,
+                                                num_scheduled_tokens,
+                                                scheduler_output))
```
xuechendi, line 2073: What is the reason we still keep a prefill_data and decode_data for unified_attention in seperate set? This is just temporary to minimize the changes for hpu_model_runner right? We will unify prefill and decode in 1.24, right?

madamczyk-intel: Yes exactly. This is left as is to minimize changes as doing 'full' mixed batches would require a lot of changes to runner. On the side note it will also enable us to simplify batch creation significantly as we wouldn't need to sort the samples + other things we need to do now.

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -1325,6 +1348,9 @@ def _align_and_pad_mrope_positions(self, req_ids: list[str],
                 dst_start += target_len
         return mrope_position_tensor
 
+    def _skip_bucketing(self, seq_lens, num_blocks):
+        return (len(seq_lens), 0, 0)
```
adobrzyn, line 1352: Why `return (len(seq_lens), 0, 0)` and not using `num_blocks` for third dim?

PR #141, opened 2025-09-05
Enable embedding feature

slokesha:
This PR adds support to hpu_model_ruuner to execute pooling models. 
Note : Warm up is not yet enabled for pooling.


adobrzyn (approved):
Doesn't crash now on warmup, thanks :) LGTM

PR #150, opened 2025-09-10
Enable interleaved sliding window for gemma3

jiminha:
Enable interleaved sliding window for gemma3    
    - Add prefill mask creation based on sliding_window
    - Add decode window_block operation for sliding_window
    - Modify ops.py to support FusedSDPA with window_size
    - Port implementation from v0 Gemma3 code
    - Add prefix-prefill block calculation with sliding_window

`PT_HPU_LAZY_MODE=1 VLLM_SKIP_WARMUP=True  lm_eval --model vllm   --model_args pretrained=google/gemma-3-4b-it,tensor_parallel_size=1,distributed_executor_backend=mp,trust_remote_code=true,max_model_len=4096,dtype=bfloat16,max_num_seqs=128   --tasks gsm8k     --batch_size 128 --num_fewshot 5`

|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6778|±  |0.0129|
|     |       |strict-match    |     5|exact_match|↑  |0.6755|±  |0.0129|


xuechendi:
Since this Feature is pushed to 1.24 per previous communication, let's hold and leave more review bandwidth for 1.23 features

vllm_gaudi/attention/backends/hpu_attn.py
```diff
@@ -535,9 +535,23 @@ def forward(
...
 
-            if self.sliding_window \
-               and attn_metadata.window_attn_bias is not None:
-                attn_bias = attn_metadata.window_attn_bias
+            common_args = self.common_attention_args(block_list, key_cache, value_cache, attn_metadata.block_size)
+
+            if self.sliding_window and seq_len > self.sliding_window:
+                if hasattr(attn_metadata, 'window_attn_bias') and attn_metadata.window_attn_bias is not None:
+                    attn_bias = attn_metadata.window_attn_bias
+                else:
```
xuechendi, line 543: please add a comment here: so people will know this path is to use FusedSDPA and line 542 is for using pytorch SDPA

vllm_gaudi/extension/features.py
```diff
@@ -35,6 +35,11 @@ def get_user_flags():
         # Non-vllm flags that are also important to print
         Env('EXPERIMENTAL_WEIGHT_SHARING', str),
         Env('PT_HPU_WEIGHT_SHARING', str),
+
+        # Sliding window flags
+        Env('PT_HPU_SDPA_QKV_SLICE_MODE_FWD', boolean),
+        Env('PT_HPU_SDPA_BC_FACTOR', int),
```
xuechendi, line 41: what is BC_factor?

vllm_gaudi/extension/features.py
```diff
@@ -35,6 +35,11 @@ def get_user_flags():
...
+        Env('VLLM_FUSEDSDPA_SLIDE_THLD', int),
```
xuechendi, line 42: what is the slide threshold? 
Suggest to add some description, it is not straight forward for user to read and set.

vllm_gaudi/extension/bucketing/common.py
```diff
@@ -57,8 +57,14 @@ def initialize(self, max_num_seqs, max_num_prefill_seqs, block_size, max_num_bat
         self.fallback_seq_base_step = 32
         self.fallback_blocks_base_step = 32
 
-    ### GENERATE BUCKETS FUNCTIONS ###
+        self.use_sliding_window = get_config().PT_HPU_SDPA_QKV_SLICE_MODE_FWD
+        if self.use_sliding_window:
+            self.slice_size = get_config().PT_HPU_SDPA_BC_FACTOR if \
+                get_config().PT_HPU_SDPA_BC_FACTOR is not None else 1024
```
xuechendi, line 63: default value should be added to features.py

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -328,6 +335,13 @@ def __init__(self, model, vllm_config):
         self._rotary_prepare_cos_sin = self._get_prepare_cos_sin()
         self.unified_attn = get_config().unified_attn
         self.flatten_input = get_config().flatten_input
+        self.is_mm_optimized = is_mm_optimized(self.model)
+        self.sliding_window = vllm_config.model_config.get_sliding_window()
+        self.interleaved_sliding_window = is_interleaved(vllm_config.model_config.hf_text_config)
+        if self.interleaved_sliding_window:
+            self.use_window_sdpa = os.getenv("PT_HPU_SDPA_QKV_SLICE_MODE_FWD", "false").strip().lower() in ("1", "true")
+            self.slice_size = int(os.getenv("PT_HPU_SDPA_BC_FACTOR", "1024"))
+            self.slice_thld = int(os.environ.get('VLLM_FUSEDSDPA_SLIDE_THLD', '8192'))
```
xuechendi, line 344: why not using get_config()

vllm_gaudi/extension/features.py
jiminha: These are one of the complex_guid lib internal slice_size. We might not need to look at this more in the latest docker. Need more investigation.

vllm_gaudi/v1/worker/hpu_model_runner.py
jiminha: Removed

jiminha:
> > @xuechendi Please review again. After discussing with Libin, we decided to remove using the special kernel(FusedSDPA with window_size) support for now. This feature is only used during the initial prefill when the blocklist is none in v1, whereas in v0 it was used for all prompts. We plan to add it back later if it proves necessary for long prompt support. Additionally, some of the required environment variable checks for this kernel may no longer be needed with the latest kernel update, but this requires further investigation
> 
> How we handle long-context in this PR?

We create our own mask, _set_attn_bias_for_sliding_window

PR #170, opened 2025-09-15
WarmUp for Pooling - Embed Task

slokesha:
Enabled Warmup for Pooling

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -1176,6 +1176,7 @@ class HPUModelRunner(KVConnectorModelRunnerMixin):
             num_computed_tokens = req_data.num_computed_tokens[i]
             new_block_ids = req_data.new_block_ids[i]
             resumed_from_preemption = req_id in getattr(req_data, "resumed_req_ids", set())
+            num_output_tokens = req_data.num_output_tokens[i]
             req_state.num_computed_tokens = num_computed_tokens
 
             if not is_last_rank:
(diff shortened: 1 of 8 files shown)


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -3671,21 +3803,29 @@ def warmup_model(self) -> None:
...
+                            True, kv_caches)
+                    mem_post_decode, decode_batch_seq, decode_captured_all = \
+                        self.warmup_graphs(
+                            self.bucketing_manager.decode_buckets,
+                            False, kv_caches)
 
                 self.log_graph_warmup_summary(self.bucketing_manager.prompt_buckets, True, mem_post_prompt)
-                self.log_graph_warmup_summary(self.bucketing_manager.decode_buckets, False, mem_post_decode)
+                if not self.is_pooling_model:
+                    self.log_graph_warmup_summary(self.bucketing_manager.decode_buckets, False, mem_post_decode)
```
xuechendi, line 3828: why this doesn't work for pooling model?

xuechendi:
PR looks good to me overall, since the warmup is currently not being tested, may you update the test here by removing skip_warmup with small range of buckets to test embedding warmup?
https://github.com/vllm-project/vllm-gaudi/blob/main/tests/full_tests/ci_gsm8k_tests.sh#L225

vllm_gaudi/v1/worker/hpu_model_runner.py
slokesha: pooling/embedding doesn’t go through the decode path. So no need to warmup decode buckets

PR #171, opened 2025-09-15
fix block bucket size for DP+contiguous PA

wuxun-zhang:
This can be reproduced when increasing input reqs.


xuechendi:
> @xuechendi @kzawora-intel Please help review. BTW, I saw in our CI tests, contiguous PA is usually disabled, is it intended or we can add test for it?

ContiguousPA was causing acc issue which seems be resolved recently, Yes, enable contiguousPA will needed for CI.

PR #172, opened 2025-09-15
[CI FIX]Fix issue introduced by upstream PR #23974

xuechendi:
[PR #23974](https://github.com/vllm-project/vllm/pull/23974) updated the vllm.v1.core.kv_cache_utils.get_kv_cache_configs api.


xuechendi:
PD is failed on recent upstream change, 
https://github.com/vllm-project/vllm/pull/24745
will fix it in following PR

PR #175, opened 2025-09-15
[Benchmark Fix] Fix recursive import and module update when VLLM_WEIGHT_LOAD_FORCE_SYNC is ON

xuechendi:


xuechendi:
oh, realized this will fail TP, now looking for solution as #173

PR #176, opened 2025-09-15
[SW-240630] Qwen3-30B-MoE: Flatten post-attn seqs and restore model output shape

attafosu:
- After [#24772](https://github.com/vllm-project/vllm/pull/24772) an assertion on input shape in `Qwen3MoeSparseMoeBlock`  exposes a design compatibility [issue #24806](https://github.com/vllm-project/vllm/issues/24806): attn output is expected to flatten `batch_size x seqlen`
- This PR aligns the attention output as such, and restores the output shape of the forward pass afterwards


xuechendi:
@attafosu , you also need to bring Qwen test back

attafosu:
> @attafosu , you also need to bring Qwen test back

Right. I've added it now

vllm_gaudi/attention/backends/hpu_attn.py
```diff
@@ -584,6 +587,12 @@ def forward(
                                                       **self.common_attention_args(block_list, key_cache, value_cache,
                                                                                    attn_metadata.block_size))
         # Reshape the output tensor.
+        # NOTE (attafosu): Need to selectively flatten the batch and seq dims.
+        # Some modules require 1D/2D attn outputs, e.g Qwen3MoeSparseMoeBlock
+        # Re: https://github.com/vllm-project/vllm/issues/24806
+        # Doing this by default for all models breaks post-attn ops for other models, e.g facebook/opt-125m.
+        if COLLAPSE_ATTENTION_OUTPUT:
```
xuechendi, line 594: Sorry, thought I replied but actually did not, this is a little too tricky, can we automatically decide if we want to collapse or not instead of using flags?

xuechendi: If we do want to use flags, can we use in hpu_model_runner as previous commit did.
Also flags should be added to extensions/features.py so it can be print out in the log

attafosu: Yeah looks cleaner in the extensions/features. Done

xuechendi (approved):

PR #178, opened 2025-09-16
Fix sed command in README

xwu-intel:
the command sed '/^[torch]/d' requirements/build.txt instructs sed to delete any line in the file requirements/build.txt that starts with the character t, o, r, c, or h.

However, it's likely intended to match the literal string "torch" at the beginning of the line. The correct regular expression for that would be ^torch.


xuechendi:
^[torch] works for current CI, please check workflow files in .github folder for CI docker build

xwu-intel:
> ^[torch] works for current CI, please check workflow files in .github folder for CI docker build

It works since there is only one item match the pattern in the requirement file. If you have another line start with 't' or 'o', it will be deleted as well, leaving a bug which is hard to discover.

PatrykWo:
@xwu-intel it's already implemented. Can we close as done?

PR #182, opened 2025-09-16
Fix swap in defragmentator

kamil-kaczor:


vllm_gaudi/extension/defragmentation.py
```diff
@@ -95,6 +95,12 @@ def use_block(self, block_id: int):
         num_refs = self.used_blocks.get(block_id, 0) + 1
         self.used_blocks[block_id] = num_refs
 
+    def swap_block(self, block_used: int, block_free: int):
+        """ Move ref-count to the new block """
+        num_refs = self.used_blocks[block_used]
+        del self.used_blocks[block_used]
+        self.used_blocks[block_free] = num_refs
+
```
madamczyk-intel, line 103: I guess it would be better to actually swap the refs. There's a hidden assumption here that we only swap a used block with a free_block and it won't work if we accidently pass the block ids in the wrong order. Something like:

```python
def get_ref_count(block_id):
    return self.used_blocks.get(block_id, 0)

def set_ref_count(block_id, ref_count):
    if ref_count <= 0:
        del self.used_blocks[block_id]
    else:
        self.used_blocks[block_id] = ref_count

def swap_refs(self, block_a, block_b):
    a_refs = self.get_ref_count(block_a)
    b_refs = self.get_ref_count(block_b)
    self.set_ref_count(block_a, b_refs)
    self.set_ref_count(block_b, a_refs)
```
This way get/set functions could be also used in use_block and free_block.

WDYT?

madamczyk-intel (approved):

adobrzyn (approved):

PR #187, opened 2025-09-17
Add Plugin V1 specific recipe changes

nngokhale:
Calculate prompt graphs and decode graphs as per latest 3d bucketing.
Add max_num_batched_tokens column and set as per v1 default
Set max_num_prefill_seqs as per v1 default
Increased 70B mem per graph setting.
Allow gpu_mem_utilization to be changed by user.

Tested on default model len and tp of validated models.

Known issues: 
New ctx_cfg (1,1,max_model_len//block_size) causes huge number of buckets  at long context.
New 3d prefill graphs are bigger in size than 2d, so mem per graph need adjustment.
Increase in graphs due ctx_cfg and increase in graph size limits long contexts supported due to memory used.


PatrykWo:
That change broke docker compose.

PatrykWo:
Tested. Working OK. Please cherry-pick to v0.10.2_next.

sys-hab-pt-service:
Only codeowners and testowners can request to run Gaudi tests. Contact list: kzawora-intel, xuechendi, mswiniarsk, adobrzyn, mgawarkiewicz-intel, vivekgoe, afierka-intel, michalkuligowski, iboiko-habana, PatrykWo, kamil-kaczor, kfojcik-intel, ksmusz, wuxun-zhang, xuechendi, attafosu, ulivne, Kacper-Pietkun, iboiko-habana, jkaniecki

PR #188, opened 2025-09-17
Specify output tensor in matmul_qk

kdamaszk:
Some models (like Qwen) require softmax to be executed in fp32 precision. In the past, we were explicitly casted `matmul_qk` output. Such solution requires `mark_step` to ensure cast will be executed, which breaks graph and could affect performance.
This PR provides solution with specifying output tensor in `torch.matmul`, so we can avoid `mark_step`.

Original idea: https://github.com/HabanaAI/vllm-hpu-extension/commit/ad671aa71efcfbdc31c55115c39824ce79c75d3d

### DIFF
--- a/vllm_gaudi/extension/ops.py
+++ b/vllm_gaudi/extension/ops.py
@@ -52,6 +52,14 @@ def block2batch(tensor, block_mapping, matmul_op=torch.matmul):
     return b2b_impl(tensor, block_mapping.t(), matmul_op)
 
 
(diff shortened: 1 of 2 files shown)


mswiniarsk (approved):

kdamaszk:
This PR requires change in neural-compressor to support output tensor in torch,matmul

michalkuligowski:
https://github.com/vllm-project/vllm-gaudi/pull/571

PR #191, opened 2025-09-17
V0.10.2 docker updates / benchmark serving section

PatrykWo:


.cd/benchmark/benchmark_defaults.yaml
```diff
@@ -16,7 +16,7 @@ model_text:
     - Qwen/Qwen2.5-7B-Instruct
     - ibm-granite/granite-8b-code-instruct-4k
     - ibm-granite/granite-20b-code-instruct-8k
-  DATASET: /workspace/vllm/benchmarks/sonnet.txt
+  DATASET: /workspace/vllm-project/benchmarks/sonnet.txt
```
xuechendi, line 19: why the path is vllm-project?

PR #194, opened 2025-09-18
Fix DP dummy run crash for P/D 

wuxun-zhang:
- Skip kvtransfer load/save for dummy run
  - For dummy run, `SchedulerOutput` is created without `kv_connector_metadata`
- Fix real prefill batch calculation


xuechendi:
@wuxun-zhang , in what case it fails P/D?

xuechendi (approved):

PR #246, opened 2025-09-24
Validate if neighbor buckets are valid

kdamaszk:
Currently we are validating if buckets are valid and after that we are adding also neighbor buckets without checking if they are also below the tokens limit. It means that even if we refuse bucket that exceeding tokens limit, it could be still added during smaller bucket selection.
This PR fixes it, by validating also neighbor buckets


kdamaszk:
Closing this, @adobrzyn will prepare a more sophisticated solution

PR #284, opened 2025-09-28
Fix DP dummy run cfg

wuxun-zhang:


wuxun-zhang:
@xuechendi Please take a look, it's related to changes from https://github.com/vllm-project/vllm-gaudi/pull/157, but not sure why it's not caught by CI.

wuxun-zhang:
@xuechendi This PR can fix the issue previously reported.
BTW, I also tested with lastest vllm commit, it breaks our DP functionality again. I prepared another patch for this (https://github.com/vllm-project/vllm-gaudi/pull/285). Merging this fix separately, or combining with PR#285, what's you preference?

xuechendi (approved):
LGTM

PR #285, opened 2025-09-28
Support sequence parallel MOE after upstream #24982

wuxun-zhang:
After https://github.com/vllm-project/vllm/pull/24982 merged, sequence parallel MOE will be turned on when `enable_expert_parallel=True`, `tp_size > 1` and `dp_size > 1`. Since for Gaudi, there is no choice for `VLLM_ALL2ALL_BACKEND`, we can not easily bypass it. So this PR aims to support the feature.

```python
class ParallelConfig:

  @property
    def use_sequence_parallel_moe(self) -> bool:
        return (envs.VLLM_ALL2ALL_BACKEND
                in ("allgather_reducescatter", "naive",
                    "deepep_high_throughput", "deepep_low_latency")
                and self.enable_expert_parallel
                and self.tensor_parallel_size > 1
                and self.data_parallel_size > 1)

```

Update:
No hard requirement on https://github.com/vllm-project/vllm/pull/25828


wuxun-zhang:
> Please try with add GraniteMOE to flatten input_ids to 1D.

Thanks, it works. Just updated.

PR #288, opened 2025-09-29
[CI][GHA] List delta between last stable commit in hourly

kzawora-intel:
requires https://github.com/vllm-project/vllm-gaudi/pull/287


kzawora-intel:
@adobrzyn will take over this PR

PR #316, opened 2025-10-06
Port: [SW-240630] Qwen3-30B-MoE: Flatten post-attn seqs and restore model o…

adobrzyn:


xuechendi:
Duplicated PR to: https://github.com/vllm-project/vllm-gaudi/pull/212

PR #330, opened 2025-10-07
Add torch

cabelo:
We really need this command, I deleted everything and retested. You need to use add torch in the tests with gaudi2 and gaudi3.


README.md
```diff
@@ -38,6 +38,7 @@ Learn more:
     cd vllm
     git checkout $VLLM_COMMIT_HASH
     pip install -r <(sed '/^[torch]/d' requirements/build.txt)
+    pip install torch --index-url https://download.pytorch.org/whl/cpu
```
michalkuligowski, line 41: This installs the cpu torch version, Im afraid habana backend will be overwritten then

PR #333, opened 2025-10-07
Fix neighboring buckets

adobrzyn:
Adjust positions_cpu to run configs with maximum prompt size (now warmup starts with 1 :) )


michalkuligowski (approved):

adobrzyn:
Thats a bad PR

PR #345, opened 2025-10-08
Install pytorch from habanalabs-installer

cabelo:
We really need this command, I deleted everything and retested. You need to use add torch in the tests with gaudi2 and gaudi3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,10 @@ Learn more:
     cd vllm
     git checkout $VLLM_COMMIT_HASH
     pip install -r <(sed '/^[torch]/d' requirements/build.txt)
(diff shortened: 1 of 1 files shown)


adobrzyn:
HI @cabelo is this PR still needed?

cabelo:
> HI @cabelo is this PR still needed?
In the projects I used without Docker and with the Gaudi accelerator, I needed to use the commands below.

PatrykWo:
@cabelo The Gaudi driver installation is not within the scope of the framework. It is also not related to the vLLM documentation. From the vLLM perspective, we assume that the HPU device is already visible on the host before vLLM is installed.

PR #359, opened 2025-10-08
Fix issue with async_scheduling when dealing with chunked input

tianmu-li:
When dealing with chunked prompt (input sequence length > max_num_batched_tokens), sometimes there is no output token due to the chunked prompt, but the scheduler expects one. This PR addresses this issue and aligns the behavior with gpu_model_runner.


piotrbocian (dismissed):

PatrykWo:
@piotrbocian the fix is for blocker, Please approve.

piotrbocian (approved):

PR #376, opened 2025-10-09
defrag: Optimize tail-compaction to reduce memcpy traffic and preserve locality

yafshar:
### What this PR does

Replaces the existing **greedy high-low swap** defragmentation algorithm with a **tail-compaction** strategy

1. Uses **ratio-based trigger** (`frag_ratio = max_phys / num_used`) vs `VLLM_DEFRAG_RATIO_LIMIT` (default 1.5).
2. Victims and holes paired by **physical ID** for maximum tail compaction:
   - Victims sorted **descending by physical ID**.
   - Free holes sorted **ascending by physical ID**.
   - Skips invalid pairs with `continue` instead of stopping early.
3. Avoids full scans; focuses on **tail region only** → complexity closer to **O(#tail_blocks)**.
4. Copies **significantly fewer KV-cache blocks per pass** (≤ half compared to original).
5. Adds **configurable minimum swap threshold** (`VLLM_DEFRAG_MIN_SWAPS`, default 4) to skip micro-defrags.
6. Retains **dynamic padding** for graph stability (`pad = 2**(len(to_swap)-1).bit_length()`).

### Why we need it

**Observation**  
Prefix cache hits are **heavily skewed to low logical block IDs** (0,1,2…).  
In the original algorithm, a formerly “hot” logical block (e.g., 3) could end up at physical block 900 after defrag.  
When the scheduler later reuses block 3, we create a **new hole at 900** and trigger **yet another defrag** – death by a thousand swaps.

**Goal**  
Reduce repeated defrags and memory churn by:
- Keeping compaction focused on the tail region.
- Avoiding unnecessary moves of hot logical blocks.
- Maintaining stable graph execution for Gaudi HPU.

### Original (main) vs Optimized (this PR) 

| Aspect                | Original (main)                | Optimized (this PR)                     |
|-----------------------|----------------------------------------------------|---------------------------------------------------------|
| Trigger condition     | Absolute threshold (`VLLM_DEFRAG_THRESHOLD`)      | Ratio-based (`frag_ratio > VLLM_DEFRAG_RATIO_LIMIT`)   |
| Victim ordering       | Logical ID ascending                              | Physical ID descending                                 |
| Hole ordering         | Static padded list                                | Physical ID ascending                                  |
| Pairing logic         | Stops on first invalid pair (`break`)             | Skips invalid pairs (`continue`)                      |
| Minimum swaps         | Not enforced                                      | Configurable (`VLLM_DEFRAG_MIN_SWAPS`, default=4)      |
| Padding               | Static thresholds                                 | Dynamic power-of-two padding for graph stability       |
| Complexity            | Full scan (`O(U+F)`)                              | Tail-focused (`O(#tail_blocks)`)                      |

### Algorithm in one sentence

Compact only the tail of physical memory using ratio-based trigger and physical pairing for maximum efficiency.

### Performance numbers (meta-llama/Llama-3.1-8B-Instruct, input=1024, output=1024, batch=170)

**Main branch (baseline):**
- Lazy mode + graph off:  
  Median iteration time ≈ **49.28 s**, throughput ≈ **3532 tok/s**  
  Swap events: **38–46**, sum of blocks swapped: **3586–4191**
- torch.compile + graph off:  
  Median iteration time ≈ **49.29 s**, throughput ≈ **3531 tok/s**  
  Swap events: **44–56**, sum of blocks swapped: **3829–5597**

**Optimized PR:**
- Lazy mode + graph off:  
  Median iteration time ≈ **49.23 s**, throughput ≈ **3535 tok/s**  
  Swap events: **11**, sum of blocks swapped: **1421**
- torch.compile + graph off:  
  Median iteration time ≈ **49.21–49.22 s**, throughput ≈ **3536–3537 tok/s**  
  Swap events: **14–16**, sum of blocks swapped: **1700–1717**
- torch.compile + graph on:  
  Median iteration time ≈ **49.21 s**, throughput ≈ **3537 tok/s**  
  Swap events: **11–13**, sum of blocks swapped: **1250–1694**

**Summary:**  
- **Swap events reduced by ~70–80%** (from 38–56 down to 11–16).  
- **Blocks swapped reduced by ~65–75%** (from 3.5K–5.5K down to ~1.2K–1.7K).  
- **Throughput slightly improved** (≈ +3–5 tok/s).  
- **Iteration time stable** (no regression).

**Note:** Throughput remains nearly unchanged because defragmentation is not in the critical path for token generation. The main bottleneck is model compute (attention/matmul), so reducing KV-cache copy volume improves fragmentation stability and memory efficiency but does not significantly affect per-token latency.

**Future improvement ideas:**  
- Use adaptive thresholds based on workload patterns
- Use batched gather/scatter kernels or async streams to overlap swaps with compute
- Trigger defrag proactively during idle cycles

### Compatibility

- No public API change 

### Early-exit heuristic

We skip runs with fewer than **VLLM_DEFRAG_MIN_SWAPS** (default = 4) swaps to avoid micro-copy overhead.  
This threshold is configurable and can be tuned if profiling shows a different sweet spot.

### DIFF
--- a/tests/unit_tests/test_defragmentation.py
+++ b/tests/unit_tests/test_defragmentation.py
@@ -195,7 +195,7 @@ class TestOnlineDefragmenter:
         # (max_phys / num_used) is <= frag_limit.
         # Here we use contiguous blocks 1..N so frag_ratio = 1.0,
         # which is <= default frag_limit (1.5).
-    
+
         # Use a dense, contiguous allocation: blocks 1..19
         for i in range(1, 20):
             defragmenter.use_block(i)


yafshar:
@ksmusz I updated the PR. Would you please review this. Thanks

yafshar:
After benchmarking, and some debugging, I am concluding that the tail-compaction algorithm is slower than the bidirectional approach in long-running scenarios due to O(max_phys) vs O(U log U) complexity "U" represents the number of used blocks (len(used_blocks)). The performance gap grows as max_phys increases over time while used_blocks remains constant.
 
 - Key finding: Tail-compaction only provides value when prefix caching is enabled & the system requires stable logical-to-physical block mapping for cache coherence. In all other cases, the bidirectional algorithm is faster, simpler, and proven.
 
- Recommendation: Default to the current bidirectional defragmentation. If prefix caching becomes a priority and we observe cache invalidation issues from block reordering, we can revisit a specialized version.
 
Closing my PR to avoid premature optimization overhead.

PR #400, opened 2025-10-13
Unified Attention - batch preparation rewrite

kzawora-intel:
This PR introduces a persistent batch optimization for unified attention, as well as full rewrite from PyTorch to NumPy. Effectively, it reduces the batch preparation time by 6-15x across the board.

### DIFF
--- a/vllm_gaudi/extension/unified_batch.py
+++ b/vllm_gaudi/extension/unified_batch.py
@@ -13,15 +13,15 @@ logger = init_logger()
 @dataclass
 class UnifiedBatch:
     req_ids_cpu: list[str]
-    token_ids: torch.tensor
-    token_positions: torch.tensor
-    new_token_positions_cpu: torch.tensor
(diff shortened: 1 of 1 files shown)


vllm_gaudi/extension/unified_batch.py
```diff
@@ -0,0 +1,329 @@
...
+from typing import Optional, Callable, Union
+from vllm_gaudi.extension.logger import logger as init_logger
+
+logger = init_logger()
+
+
+@dataclass
+class UnifiedBatch:
+    req_ids_cpu: list[str]
+    token_ids: torch.tensor
```
yafshar:
```suggestion
    token_ids: torch.Tensor
```

I think you should update `torch.tensor` -> `torch.Tensor` in the whole commit
`torch.tensor` is the factory function, not the type and the type annotation should be `torch.Tensor` (capital T).

afierka-intel (approved):

PR #409, opened 2025-10-15
Port: [Bugfix] Fix min linear decode value (#391)

adobrzyn:
Currently out of the box linear doesn't work, this should help

---------


michalkuligowski (approved):

wpyszka (approved):
fix for a bug, approved

PR #411, opened 2025-10-15
[NIXL][BUGFIX][Gaudi2Gaudi accuracy] use 4d kv_cache for nixl_connector KV register and update host_buffer accordingly

xuechendi:
Implement for SW-242433

** Root cause for accuracy issue **

nixl_connector register_kv_cache assume any input kv_caches are with 4D or 5D. 
`[2, num_blocks, block_size, num_kv_heads, head_size]` or `[num_blocks, block_size, num_kv_heads, head_size]`

However, HPU KV cache is with 3D tuple: `Tuple([num_blocks*block_size, num_kv_heads, head_size], ...)`

=> Different KV layout leads to incorrect num_blocks and data_ptr calculation in nixl_connector => leads to wrong data copied.

** Solution **

1. create a new KV_caches_4D dict for nixl_conenctor. This 4D KV_caches is a reference to original KV_cache with 4D view. (Same memory address)
2. Fix `habana_frameworks.torch.utils.experimental._data_ptr` incapability of fetching address on view-tensor by using global map to map virtual to physical addr.
3. add a new TupleTensor Class which treat tuple as Tensor to return shape, device, dtype

** validation **

Tested with both Gaudi2Gaudi and Gaudi2CPU2Gaudi on "Qwen/Qwen3-0.6B",  "deepseek-ai/DeepSeek-V2-Lite-Chat"

All 4 cases get expected accuracy.

### DIFF
--- a/vllm_gaudi/platform.py
+++ b/vllm_gaudi/platform.py
@@ -218,9 +218,10 @@ class HpuPlatform(Platform):
         if isinstance(dst_cache, tuple):
             _src_cache = src_cache[:, src_block_indices]
             for i in range(len(dst_cache)):
-                dst_cache[i].index_put_((dst_block_indices, ), _src_cache[i].to(dst_cache[i].device))
+                dst_cache[i].index_copy_(0, dst_block_indices, _src_cache[i].to(dst_cache[i].device))
         else:
-            dst_cache.index_put_((dst_block_indices, ), src_cache[src_block_indices].to(dst_cache.device))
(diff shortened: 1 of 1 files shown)


xuechendi:
@xinyu-intel  @hsubramony  @kzawora-intel 
Please help to review this PR to fix nixl_connector accuracy issue.

Meanwhile, I now take most use of nixl_connector original codes to avoid codes copy and override, 
including:
- use original initialize_host_xfer_buffer
- use original copy_kv_blocks
- use original register_kv_caches

PR #415, opened 2025-10-17
reuse DP allgather tensor across layers

wuxun-zhang:
This is to reduce cached memory size for DP dispatch/combine when HPU graph enabled.


xuechendi (approved):
LGTM

PR #421, opened 2025-10-18
[NIXL]Enable prefill TP < Decode TP with host_buffer

xuechendi:
SW-242361

Change DecodeTP=2, PrefillTP=1 for test

Issue:
 - HPU attn is using NHD, and vllm upstream only support prefill / decode heterogeneous TP with HND.

Solution:
 - init:
    hpu_attn with NHD -> host_buffer with HND

 - copy device to host:
    permute kv for req -> copy to host buffer

 - nixl_connector transfer host KV with HND + TP_ratio support

 - copy host to device
    permute kv for req -> copy to device

=====

Validated, accuracy is good

--- 
FYI, no need change for MLA (Deepseek)

### DIFF
--- a/tests/models/test_initialization.py
+++ b/tests/models/test_initialization.py
@@ -8,7 +8,7 @@ from transformers import PretrainedConfig
 
 from vllm import LLM
 from vllm.engine.llm_engine import LLMEngine as V0LLMEngine
-from vllm.utils import GiB_bytes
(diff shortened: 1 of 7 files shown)


xuechendi:
@hsubramony @xinyu-intel , please help to take a look.
This PR is aimed to enable prefill TP < Decode TP with CPU offloading path. CI is also updated using prefill TP=1 and Decode TP=2.

PR #331, opened 2025-10-07
[SW-241908] Omit all prompt buckets that exceed max_num_batched_tokens

skavulya:
[SW-241908] Fixes regression in tests due to invalid buckets generated if VLLM_PROMPT_BS_BUCKET_MAX is set, and the number of tokens in prefill batch exceeds the max_num_batched_tokens. The regression is associated with https://github.com/vllm-project/vllm-gaudi/pull/224   

This fix checks that the number of tokens in both the current and next token bucket does not exceed max_num_batched tokens, and resolves "ValueError: operands could not be broadcast together with shape" exception

`VLLM_CONTIGUOUS_PA=false VLLM_DECODE_BLOCK_BUCKET_MAX=512 VLLM_USE_V1=1 VLLM_PROMPT_BS_BUCKET_MAX=16 python  redhat-int4-plugin.py`

(EngineCore_DP0 pid=352574)   File "/vllm-gaudi/vllm_gaudi/v1/worker/hpu_model_runner.py", line 4051, in warmup_model
(EngineCore_DP0 pid=352574)     self.warmup_graphs(
(EngineCore_DP0 pid=352574)   File "/vllm-gaudi/vllm_gaudi/v1/worker/hpu_model_runner.py", line 3659, in warmup_graphs
(EngineCore_DP0 pid=352574)     self._prepare_dummy_scenario(prompt_cfg, decode_cfg)
(EngineCore_DP0 pid=352574)   File "/vllm-gaudi/vllm_gaudi/v1/worker/hpu_model_runner.py", line 3897, in _prepare_dummy_scenario
(EngineCore_DP0 pid=352574)     self._execute_dummy_scenario(requests, scheduled_tokens)
(EngineCore_DP0 pid=352574)   File "/vllm-gaudi/vllm_gaudi/v1/worker/hpu_model_runner.py", line 3928, in _execute_dummy_scenario
(EngineCore_DP0 pid=352574)     self.execute_model(sched_output, warmup_mode=True)
(EngineCore_DP0 pid=352574)   File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 120, in decorate_context
(EngineCore_DP0 pid=352574)     return func(*args, **kwargs)
(EngineCore_DP0 pid=352574)   File "/vllm-gaudi/vllm_gaudi/v1/worker/hpu_model_runner.py", line 2914, in execute_model
[...]

PR #439, opened 2025-10-21
[v0.11.0] CP Update not_over_max_model_len filter to fix warmup perf regression…

kamil-kaczor:
Fix performance regression caused by missing warmup buckets associated with https://github.com/vllm-project/vllm-gaudi/pull/331


iboiko-habana (approved):

wpyszka (approved):
bugfix, approved

PR #440, opened 2025-10-21
[v0.10.2] CP Update not_over_max_model_len filter to fix warmup perf r…

kamil-kaczor:
Fix performance regression caused by missing warmup buckets associated with https://github.com/vllm-project/vllm-gaudi/pull/331


wpyszka (approved):
bugfix, approved

PR #457, opened 2025-10-23
Add tests for custom operator implementation correctness

Kacper-Pietkun:
I added tests for custom ops defined in `vllm_gaudi/ops`:
- For the tests of ops that are not using cuda kernels - native ops and hpu ops are triggered for the same input and their outputs are compared
- For others tests that are using cuda kernels (so cannot be called with vllm-gaudi plugin) I created separate directory to store some predefined small tensors - weights, inputs and outputs. These tensors are too big to hardcode them in tests, however their sizes were adjusted, so all of them weight less than 3MB in total. Tensors are stored in a .safetensors format. Such tests run hpu ops with loaded inputs and weights and compare their outputs with the loaded outputs.


Kacper-Pietkun:
All of the above changes are just corrections of spelling mistakes detected by copilot

PR #467, opened 2025-10-24
fix dummy run config for P/D prefiller instance

wuxun-zhang:
When P/D disaggregation + DP used, prefiller instance should run dummy prefill run instead of decode run when certain DP rank has no work to do.


wuxun-zhang:
@xuechendi can you please take a review here? Thanks

xinyu-intel (approved):

PR #475, opened 2025-10-24
[Attention Metadata Overhaul 3/N] Add per-layer attention metadata

kzawora-intel:


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -583,14 +608,54 @@ def _maybe_wrap_in_hpu_graph(*args, **kwargs):
             *args, **kwargs)
 
 
-def subtuple(obj: object, typename: str, to_copy: list[str], to_override: Optional[dict[str, object]] = None):
+class HashableDict(dict):
+
+    def __hash__(self):
+        return hash((frozenset(self), frozenset(self.values())))
+
```
jkaniecki: Is this used somewhere ?

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -4180,6 +4255,9 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
             raise NotImplementedError("Hybrid models with more than one KV cache type are not "
                                       "supported yet.")
 
+        kv_cache_config = deepcopy(kv_cache_config)
+        self.kv_cache_config = kv_cache_config
+
```
jkaniecki: Why do we need this ?

kzawora-intel: we will need to access at least `self.kv_cache_config.kv_cache_groups` when we'll be preparing different metadata instances for different attention types - each group should have one distinct instance of attn metadata - it's how we can distinguish between chunked and non-chunked attn, or any other hybrid shenanigans. in the current state, we don't bother and just prepare one instance regardless of number of groups - i'm not sure if i'll complete it in this PR tho or the next ones, since it's already a big can of worms

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -524,8 +537,20 @@ def forward(self, *args, **kwargs):
...
+                                                     input_ids.device, self.dtype)
+            for key in kwargs['attn_metadata']:
+                if kwargs['attn_metadata'][key] is first_metadata:
+                    kwargs['attn_metadata'][key] = updated_metadata
+                else:
+                    msg = f"Different attn_metadata encountered on layer {key}. Updating it individually."
+                    logger.warning(msg)
+                    kwargs['attn_metadata'][key] = self._update_metadata(kwargs['attn_metadata'][key],
+                                                                         input_ids.size(0), input_ids.size(1),
+                                                                         input_ids.device, self.dtype)
```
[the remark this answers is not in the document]
kzawora-intel: because both are trash, i'm currently overhauling this specific part

PR #521, opened 2025-11-05
enable lmcache 

hsubramony:
enable lmcache for hpu

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -39,6 +39,7 @@ from vllm.attention.backends.abstract import AttentionType
 from vllm.attention.layer import Attention
 from vllm.attention.layer import MLAAttention
 from vllm.attention.selector import get_attn_backend
+
 from vllm.config import (VllmConfig, update_config)
 from vllm.distributed.kv_transfer import (get_kv_transfer_group, has_kv_transfer_group)
 from vllm.forward_context import set_forward_context
@@ -93,6 +94,8 @@ from vllm_gaudi.extension.ops import LoraMask as LoraMask
 from vllm.distributed.kv_transfer.kv_connector.utils import copy_kv_blocks
 from vllm.distributed.kv_transfer.kv_connector.v1.nixl_connector import NixlConnectorMetadata
 from vllm.v1.core.sched.output import GrammarOutput
+from vllm.config.multimodal import ImageDummyOptions
+from vllm.multimodal.profiling import MultiModalProfiler
 
 if TYPE_CHECKING:
     import xgrammar as xgr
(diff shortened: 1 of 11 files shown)


michalkuligowski (commented):
Can you add small UT to plugin CI?

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -3219,9 +3227,6 @@ def sample_tokens(self, grammar_output: "GrammarOutput | None") -> ModelRunnerOu
         #if not has_kv_transfer_group():
         #    assert not (num_prefills > 0 and num_decodes > 0)
         # skip kv_connector if dummy run
-        if not warmup_mode:
-            with set_forward_context(None, self.vllm_config):
-                self.maybe_setup_kv_connector(scheduler_output)
```
xuechendi: why move it into `_execute_model_generic`

vllm_gaudi/v1/worker/hpu_worker.py
```diff
@@ -74,6 +74,7 @@ def __init__(
 
         self.local_rank = local_rank
         self.rank = rank
+        self.parallel_config.rank = rank
```
xuechendi: What is default value?

vllm_gaudi/v1/worker/hpu_worker.py
```diff
@@ -328,8 +329,13 @@ def init_worker_distributed_environment(
...
     """Initialize the distributed environment."""
-    init_distributed_environment(parallel_config.world_size, rank, distributed_init_method, local_rank, backend='hccl')
-    ensure_model_parallel_initialized(parallel_config.tensor_parallel_size, parallel_config.pipeline_parallel_size)
+    init_distributed_environment(
+        world_size=parallel_config.world_size,
+        rank=rank,
+        local_rank=local_rank,
+        distributed_init_method=distributed_init_method,
+        backend='hccl',
+    )
```
xuechendi: seems this func is stay unchanged?

vllm_gaudi/v1/worker/hpu_model_runner.py
xuechendi: I see, does that work with HPU Graph? 
Wondering can we return metadata from _execute_model_generic and used by self.maybe_setup_kv_connector？

hsubramony: we could return it ,but in our case setup kv is done before the forward() , but for other cases its ok to return i guess

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -2593,6 +2598,8 @@ def _execute_model_generic(self,
         else:
             model_event_name = 'model_executable'
         with self.profiler.record_event('internal', model_event_name):
+            with set_forward_context(attn_metadata, self.vllm_config):
+                self.maybe_setup_kv_connector(scheduler_output)
```
wuxun-zhang: `execute_model_generic` could be called multiple times when num prefill batch > 1 while `scheduler_output` keeps same, will `maybe_setup_kv_connector` work well in such case?

hsubramony: can u give me an example test ?

vllm_gaudi/v1/worker/hpu_worker.py
xuechendi: can you remove the change?

vllm_gaudi/v1/worker/hpu_model_runner.py
xuechendi: please share the profile to make sure no HPU graph overhead introduced

wuxun-zhang:
@hsubramony Looks it breaks dp test, please take a look.

xinyu-intel (approved):

PR #526, opened 2025-11-05
[Attention Metadata Overhaul 1/N] Extract metadata update to HPUAttentionMetadataProcessor 

kzawora-intel:
This PR is pretty simple - it takes all the metadata post-processing logic we do inside adapter, and yeets it from there into a separate class. This shouldn't introduce any functional changes other than a small refactor. In the next PR, I intend to remove metadata postprocessing from the adapter and do it beforehand, on CPU, but I didn't want to introduce too major changes here. 

I made this because I absolutely hated how https://github.com/vllm-project/vllm-gaudi/pull/475 ended up w.r.t. metadata postprocessing, so I'd like to gradually fix it before that PR lands.

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -390,9 +390,6 @@ class HpuModelAdapter(torch.nn.Module, KVConnectorModelRunnerMixin):
                         self.model.multi_modal_projector, \
                         disable_tensor_cache=True)
 
-    def _update_metadata(self, attn_metadata, batch_size, seq_len, device, dtype):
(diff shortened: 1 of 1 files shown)


jkaniecki (changes requested):
Overall good change with one modification proposed

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -390,6 +390,9 @@ def __init__(self, model, vllm_config):
                         self.model.multi_modal_projector, \
                         disable_tensor_cache=True)
 
+    def _update_metadata(self, attn_metadata, batch_size, seq_len, device, dtype):
```
jkaniecki: Maybe let's remove this function completely as before - just add metadata_processor.process_metadata in _compile_methods()

PR #528, opened 2025-11-05
Fix unified preemption no attr found

kamil-kaczor:
Running:
`VLLM_SKIP_WARMUP=false \
VLLM_UNIFIED_ATTN=1 \
PT_HPU_LAZY_MODE=1 \
vllm serve granite-3b-code-instruct-128k/ \
    --tensor-parallel-size 1 \
    --max-num-seqs 42 \
    --block-size 128 \
    --disable-log-stats \
    --dtype bfloat16 \
    --max-model-len 2048 \
    --async_scheduling`
and user:
`python -m vllm.entrypoints.cli.main bench serve \
    --backend vllm \
    --model granite-3b-code-instruct-128k/ \
    --tokenizer granite-3b-code-instruct-128k/ \
    --percentile-metrics ttft,tpot,itl,e2el \
    --metric-percentiles 50,90,99 \
    --ignore-eos \
    --num-prompts 42 \
    --port 8000 \
    --dataset-name random \
    --random-input-len 1024 \
    --random-output-len 1024 \
    --max-concurrency 42`
we run into resumed_req_ids has no attribute "resumed_req_ids". This is covering this case.

### DIFF
--- a/.cd/Dockerfile.rhel.tenc.pytorch.vllm
+++ b/.cd/Dockerfile.rhel.tenc.pytorch.vllm
@@ -12,7 +12,7 @@ ARG TORCH_TYPE_SUFFIX
 
 FROM ${DOCKER_URL}/${VERSION}/${BASE_NAME}/${REPO_TYPE}/pytorch-${TORCH_TYPE_SUFFIX}installer-${PT_VERSION}:${REVISION}
 
-# Parameterize commit/branch for vllm-fork checkout
+# Parameterize commit/branch for vllm-plugin checkout
 ARG VLLM_GAUDI_COMMIT=main
 # leave empty to use last-good-commit-for-vllm-gaudi
 ARG VLLM_PROJECT_COMMIT=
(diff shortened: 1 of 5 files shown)


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -1184,7 +1184,7 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> bool:
             req_state = self.requests[req_id]
             num_computed_tokens = req_data.num_computed_tokens[i]
             new_block_ids = req_data.new_block_ids[i]
-            resumed_from_preemption = req_id in req_data.resumed_req_ids
+            resumed_from_preemption = req_id in getattr(req_data, "resumed_req_ids", set())
```
michalkuligowski: It shouldnt be None here, it should come from CachedRequestData. Wont this change just cover the issue?

michalkuligowski (changes requested):

PR #529, opened 2025-11-05
Doc updates: introduction and developer guides

mhelf-intel:
I updated the homepage and developer guides. The Profiling content was taken from the Habana docs website.

Please let me know if you have any comments.

### DIFF
--- a/docs/design/plugin_system.md
+++ b/docs/design/plugin_system.md
@@ -75,7 +75,7 @@ The `register_ops()` function, defined in `vllm_gaudi/__init__.py`, registers al
 
 ```python
 def register_ops():
-    """Register custom options for the HPU platform"""
+    """Register custom operations for the HPU platform"""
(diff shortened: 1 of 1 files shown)


michalkuligowski (approved):

docs/design/plugin_system.md
```diff
@@ -137,53 +89,28 @@ def register_ops():
     import vllm_gaudi.ops.hpu_multihead_attn  # noqa: F401
 ```
 
-These custom operations are imported (not called) to register them with vLLM's operation registry, making them available for use throughout the inference pipeline.
+These custom operations are imported, not called, to register them with vLLM’s operation registry so they can be used across the inference pipeline.
```
PatrykWo: We changed the meaning of the sentence here. Custom operations are imported but they are not called.
Differently : Customer operations are not called but imported.

PR #542, opened 2025-11-06
Michalkuligowski patch 7

michalkuligowski:


vllm_gaudi/extension/features.py
```diff
@@ -51,13 +51,13 @@ def get_user_flags():
 
 def get_experimental_flags():
```
michalkuligowski: No need for thia method, it is covered by the metaclass

PR #544, opened 2025-11-07
Refactor part of spec decode structure identical to vLLM

jerrychenhf:
Currently a lot of spec decode related code were implemented directly in hpu model runner and model runner will gets piled up into complexity when more spec decode improvements and features are added.

We need to follow the same structure as vLLM implementation to put as much spec decode related code to the corresponding proposer.

This refactor will make a basic refactor following this structure idea but try touch as less as possible the detailed implementation logic so that the functionality is not impacted.

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -4571,15 +4571,15 @@ class HPUModelRunner(KVConnectorModelRunnerMixin):
                 for idx, (req_id, prompt_len, token_ids, position_ids, attn_metadata, logits_indices,
                           logits_requests) in enumerate(zip(*shallow_tuple(prefill_data))):
                     _draft_token_ids = self.propose_eagle_prefill(
-                            prefill_sampled_token_ids_tensor,
-                            hidden_states_prefills,
-                            aux_hidden_states_prefills,
-                            idx,
-                            token_ids,
-                            position_ids,
-                            attn_metadata,
-                            logits_indices,
-                        )
+                        prefill_sampled_token_ids_tensor,
+                        hidden_states_prefills,
+                        aux_hidden_states_prefills,
+                        idx,
+                        token_ids,
+                        position_ids,
+                        attn_metadata,
+                        logits_indices,
+                    )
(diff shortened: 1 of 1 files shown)


jerrychenhf:
@xuechendi  Please have a review.
Because of quite some code to support num_speculative_tokens > 1 can be added to propose function in HupEagleProposer instead of in hpu model runner, I would like the new improvements are done based the refactored code for clarity.

jerrychenhf:
@xuechendi There is no changes in the detailed logic to avoid any regression or broken on functionality. 

The eagle3 run with the same configuration as you mentioned in the JIRA with the refactored code is as following. This is identical to what you got:

================= spec_eagle3 =================
latency: 18.88802695274353
acc_counts: [13627, 0]
acc_rate: 0.7124111250522793
num_draft_tokens: 19128
num_drafts: 19128

michalkuligowski:
> I can approve the PR after eagle3 CI added and passed. Will count on @michalkuligowski for merge, since I want to hold the big refactor after our Hourly fixed (Which also introduces big change to hpu_model_runner)

we upgraded vllm commit sha after fixing hourly

xuechendi (approved):

jerrychenhf:
There is no other changes other than the CI test case. The test case should run right if the hf token is set OK in the CI. It seems the failure is because of the other things. I will try rebase with latest code and see CI result.

jerrychenhf:
The test case failed for an existing issue: assert spec_decode_metadata is not None.
It is not showing up in local runs. Fix it.

jerrychenhf:
@xuechendi  I saw there are new errors on latest when running the test of eagle3 even I fixed a old issue. These issues are not happening on a main branch at Nov 6.
I also verified that the current main branch without my refactor code has the same issues caused by the recent commits. So there are regressions which was not tracked.

jerrychenhf:
@xuechendi  I made more tests on the current main branch and I saw eagle are all broken even after applying a fix to decode_data.spec_decode_metadata None assert (refer to: https://github.com/jerrychenhf/vllm-gaudi/tree/spec-decode-main-broken). The error says the a recipe compile failure.

PR #608, opened 2025-11-21
Add VLLM_DEFRAG env to enable defrag without UA

xwu-intel:
We need to enable defrag with contiguous pa for deepseek r1. Add the env to enable defrag.


xwu-intel:
Closed as there is VLLM_DEFRAG created for defrag by default.

PR #631, opened 2025-11-25
Docs: Update to Gaudi 1.22.2 and vLLM 0.11.2

mhelf-intel:
I updated the versions and hid the unified attention feature as it will be supported in 1.24.0.

### DIFF
--- a/.cd/docker-compose.yml
+++ b/.cd/docker-compose.yml
@@ -43,5 +43,5 @@ services:
     env_file:
       - ./benchmark/benchmark_user.env
     volumes:
-      - ./logs:/root/scripts/logs
+      - /tmp/logs:/root/scripts/logs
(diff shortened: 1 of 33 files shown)


docs/features/unified_attn.md
```diff
@@ -5,7 +5,7 @@ title: Unified Attention
 
 ## Overview
 
-vLLM Hardware Plugin for Intel® Gaudi® v1.23.0 introduces a new attention backend called Unified Attention, which unifies several previous algorithms into a single implementation. As a newly added backend, it currently supports a subset of the planned features. Compared to earlier approaches, it provides the following advantages:
+vLLM Hardware Plugin for Intel® Gaudi® v1.24.0 introduces a new attention backend called Unified Attention, which unifies several previous algorithms into a single implementation. As a newly added backend, it currently supports a subset of the planned features. Compared to earlier approaches, it provides the following advantages:
```
PatrykWo: It's true when 1.24 will be public, but it's future. Should we change the sentence and explain that it is future feature that productization is targeted for 1.24?

PR #635, opened 2025-11-26
nixl: add lancher for PD workloads

xinyu-intel:
- a launcher to support mix TPxDP config.


adobrzyn:
@xinyu-intel - is this still needed?

PR #645, opened 2025-11-27
bucket: add query len 1 to prefill bucket

xinyu-intel:
Avoid the query length(1) of the prefix prefill on the decode side to be padded to the block size under PD+DP scenario.

Use case:

VLLM_EXPONENTIAL_BUCKETING=false/true VLLM_PROMPT_QUERY_BUCKET_MIN=1 on the decode side.


xuechendi (commented):
mostly looking good to me, quick question
Do we also need to update for bucketing/linear.py
https://github.com/xinyu-intel/vllm-gaudi/blob/fd61e587f195962209711bb88f2e0cccbcf67524/vllm_gaudi/extension/bucketing/linear.py#L19C35-L19C55

adobrzyn (approved):

PR #647, opened 2025-11-27
CustomOp: grouped topk

xinyu-intel:
depends on https://github.com/vllm-project/vllm/pull/29575

### DIFF
--- a/vllm_gaudi/ops/hpu_fused_moe.py
+++ b/vllm_gaudi/ops/hpu_fused_moe.py
@@ -4,11 +4,81 @@ from typing import Union
 import torch
 import vllm
 from vllm.model_executor.layers.batch_invariant import vllm_is_batch_invariant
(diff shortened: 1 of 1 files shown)


xuechendi:
is that possible to monkey patch from `vllm.model_executor.layers.fused_moe.fused_moe.grouped_topk` ?

I think we can push for the https://github.com/vllm-project/vllm/pull/29575 after, since it usually need some discussion and alignment.

xinyu-intel:
> is that possible to monkey patch from `vllm.model_executor.layers.fused_moe.fused_moe.grouped_topk` ?
> 
> I think we can push for the [vllm-project/vllm#29575](https://github.com/vllm-project/vllm/pull/29575) after, since it usually need some discussion and alignment.

https://github.com/vllm-project/vllm-gaudi/pull/708

iboiko-habana:
it is merged into https://github.com/vllm-project/vllm-gaudi/pull/735

PR #658, opened 2025-11-28
Optimize MoE via chunk settings

xinyu-intel:
add new feature `VLLM_MOE_CHUNK` and `VLLM_MOE_TOKEN_BOUNDARY`, with this, chunk_size and global_num_experts will be passed to torch.ops.hpu.mixture_of_experts for better performance. 

example, for ds_r1, we use the following setting for better performance.

  export VLLM_MOE_CHUNK="64, 128"
  export VLLM_MOE_TOKEN_BOUNDARY="2048, 4096"

### DIFF
--- a/vllm_gaudi/extension/features.py
+++ b/vllm_gaudi/extension/features.py
@@ -102,6 +102,8 @@ def get_features():
         Value('track_graph_compilation', False, env_var='PT_HPU_METRICS_GC_DETAILS', env_var_type=boolean),
         Value('use_output_tensor_in_matmulqk',
               All(VersionRange(">=1.24.0.171"), MinPackageVersion("neural_compressor_pt", "3.6")),
-              env_var_type=boolean)
+              env_var_type=boolean),
+        Value('moe_chunk', "", env_var='VLLM_MOE_CHUNK', env_var_type=list_of(int)),
+        Value('moe_token_boundary', "", env_var='VLLM_MOE_TOKEN_BOUNDARY', env_var_type=list_of(int)),
     ]
     return split_values_and_flags(features)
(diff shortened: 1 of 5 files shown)


xinyu-intel:
> codes looks good to me, agree with @yiliu30 that since the magic number gets defined multiple times, adding a global var such as "CHUNK_SIZE_SMALL"/"CHUNK_SIZE_MEDIUM"/CHUNK_SIZE_LARGE, might be good option. Also, does the magic number changes to Gaudi2, Gaudi3 and Gaudi2D?

Follow the changes in https://github.com/HabanaAI/vllm-hpu-extension/pull/396

I added two list of int based env `VLLM_MOE_CHUNK` and `VLLM_MOE_TOKEN_BOUNDARY` which can specify chunk_size for different token num.

We are using the same settings for DS-R1 PD workload on G2D and G3D. For broader workloads, it needs to be tuned case by case.

ranzhejiang:
> > codes looks good to me, agree with @yiliu30 that since the magic number gets defined multiple times, adding a global var such as "CHUNK_SIZE_SMALL"/"CHUNK_SIZE_MEDIUM"/CHUNK_SIZE_LARGE, might be good option. Also, does the magic number changes to Gaudi2, Gaudi3 and Gaudi2D?
> 
> Follow the changes in [HabanaAI/vllm-hpu-extension#396](https://github.com/HabanaAI/vllm-hpu-extension/pull/396)
> 
> I added two list of int based env `VLLM_MOE_CHUNK` and `VLLM_MOE_TOKEN_BOUNDARY` which can specify chunk_size for different token num.
> 
> We are using the same settings for DS-R1 PD workload on G2D and G3D. For broader workloads, it needs to be tuned case by case.

This pull request is a good fit for the DeepSeek R1 branch, and I believe there are no major issues with the code.

PR #673, opened 2025-12-03
Block matmul and kv_cache in dynamic quantization

HolyFalafel:
Currently disabling matmul and kv_cache in dynamic quantization mode


xuechendi:
Is it for performance reason?

PR #684, opened 2025-12-04
DP: dispatch fp8 hidden_states in INC

xinyu-intel:
depends on #680


xuechendi (approved):
LGTM, @yiliu30 may you help to review as well

xuechendi:
@yiliu30 , please help to approve if you think the fix is looking good

PR #722, opened 2025-12-16
DP: Fix for torch.compile

xinyu-intel:


wuxun-zhang (approved):

vllm_gaudi/ops/hpu_fused_moe.py
```diff
@@ -57,13 +57,14 @@ def forward_oot(
         topk_weights = topk_weights.to(x.dtype)
 
         if layer.dp_size > 1:
-            hidden_states_across_dp = get_hpu_dp_metadata().hidden_states_across_dp
+            dp_metadata = get_hpu_dp_metadata()
+            hidden_states_across_dp = dp_metadata.hidden_states_across_dp if dp_metadata is not None else None
```
michalkuligowski: if dp_size >1 the get_hpu_dp_metadata() should not return None, instead of checking if None, please add assert here

xinyu-intel: it's only available under lazy+hpugraph mode since it can reuse the memory.

vllm_gaudi/ops/hpu_fp8.py
```diff
@@ -171,14 +171,16 @@ def apply(
             topk_weights = topk_weights.to(x.dtype)
 
         if layer.dp_size > 1:
+            dp_metadata = get_hpu_dp_metadata()
             if not (has_quant_config(layer.vllm_config.model_config) and self.use_dispatch_fn):
-                hidden_states_across_dp = get_hpu_dp_metadata().hidden_states_across_dp
+                hidden_states_across_dp = get_hpu_dp_metadata(
+                ).hidden_states_across_dp if dp_metadata is not None else None
```
yafshar:
```suggestion
                hidden_states_across_dp = dp_metadata.hidden_states_across_dp if dp_metadata is not None else None
```

PR #734, opened 2025-12-17
Apply hw aligned scale

lkk12014402:
refer this implementation: https://github.com/HabanaAI/vllm-fork/blob/habana_main/vllm/model_executor/layers/quantization/compressed_tensors/schemes/compressed_tensors_w8a8_fp8.py#L91


xuechendi:
rebase after #720 merged

PR #753, opened 2025-12-23
Prefill batching logic to handle chunked prefill/prefix caching for HPU

hlin99:
Logic to handle chunked prefill/prefix caching for HPU
Due to HPU padding constraints, batching requests with existing
history (ctx != 0) causes excessive memory usage, as the entire
batch must be padded to the longest context, leading to OOM.

This patch enforces a batch size of 1 for prefill operations when
ctx != 0. Although this sacrifices some throughput in corenr cases,
it effectively eliminates the OOM risk.

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -1635,6 +1635,29 @@ class HPUModelRunner(KVConnectorModelRunnerMixin):
             return self._bucketize_2d_prompt
 
     def _can_merge_prefill_contents(self, lhs, rhs):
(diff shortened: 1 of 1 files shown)


xuechendi:
@hlin99 , hmm, after second thought, I am a little unsure how the changes impact unified attention. Will need @kzawora-intel  to check

hlin99:
> @hlin99 , hmm, after second thought, I am a little unsure how the changes impact unified attention. Will need @kzawora-intel to check

sure. i wan't aware ua goes to the same code path. if ua can overcome HPU padding nature, we can definitely split the code into ua & non-ua paths. @kzawora-intel  please advise. thx

PR #777, opened 2026-01-05
[GAUDISW-244336] Add missing long ctx prompt buckets (#739)

kfojcik-intel:
To generate all missing buckets for long ctx, first we generate ctx range up to max ctx for all queries;
later we generate additional buckets for max ctx per query.

---------

Cherry-pick of https://github.com/vllm-project/vllm-gaudi/pull/739


wpyszka (approved):
lgtm, approved for 0.13

PR #779, opened 2026-01-06
Migrate `INCConfig` for HPU

yiliu30:
As part of https://github.com/vllm-project/vllm/pull/31716, the `INCConfig` in vllm-project/vllm will be used for GPU/CPU. However, we still need a placeholder for the INC path in the plugin.

cc @hshen14 @thuang6 @kzawora-intel @xuechendi


iboiko-habana:
run_deepseek_v2_inc_dynamic_tp2_test is failed because of CI issues. Test case will be disabled ASAP and fix after that

PR #786, opened 2026-01-07
create HPUConv3D class, which replaces unfold with view.

skaulintel:
Replace unfold with view in Conv3dLayer, as it fallbacks to cpu otherwise. Needed for Qwen2.5-VL-7B

### DIFF
--- a/vllm_gaudi/ops/hpu_conv.py
+++ b/vllm_gaudi/ops/hpu_conv.py
@@ -3,28 +3,6 @@ import torch.nn.functional as F
 from vllm.model_executor.layers.conv import Conv2dLayer, Conv3dLayer
 
 
-@Conv3dLayer.register_oot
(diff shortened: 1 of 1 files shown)


vllm_gaudi/ops/hpu_conv.py
```diff
@@ -1,6 +1,59 @@
...
+            x,
+            self.weight,
+            self.bias,
+            stride=self.stride,
+            padding=self.padding,
+            dilation=self.dilation,
+            groups=self.groups,
+        )
+        return x
+
```
jiminha: I don't think you need to override other apis which you don't change any code..  just need to override foward_mulmat.

iboiko-habana (commented):
Please move class HPUConv3dLayer after class HPUConv2dLayer, as it is done in upstream

PR #788, opened 2026-01-08
Set device according to local rank

yangulei:
### Motivation
For a typical node with 8xGaudi2E HPUs, the devices are break into two groups with 4 HPUs connected with top board each. Current random mapping between `local_rank` and `module_id` will cause HCCL failure for `world_size>4` cases.

### Changes
- Set device according to local rank.
- Use `pyhlml` to set `HABANA_VISIBLE_MODULES` to available modules. This is necessary if multiple cases with `world_size=1/2/4` wants to run on the same node simultaneously or the available `module_ids` are not start with 0.

### DIFF
--- a/.cd/Dockerfile.rhel.ubi.vllm
+++ b/.cd/Dockerfile.rhel.ubi.vllm
@@ -6,7 +6,14 @@ ARG ARTIFACTORY_URL="vault.habana.ai"
 ARG SYNAPSE_VERSION=1.23.0
 ARG SYNAPSE_REVISION=695
 ARG BASE_NAME=rhel9.6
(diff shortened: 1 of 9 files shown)


yafshar:
I’m not sure this approach is entirely safe. There’s a potential race condition here: if devices become busy after the availability check but before they’re actually used, the assumption of idle state could fail. This scenario is especially likely when running vllm-gaudi in Kubernetes with multiple pods scheduled on the same node (with user not using device plugins and resource limits)

iboiko-habana:
run_deepseek_v2_inc_dynamic_tp2_test is failed because of CI issues. Test case will be disabled ASAP and fix after that

PR #791, opened 2026-01-08
Unified Attention - multi-step low-level profiling

kzawora-intel:
This PR adds low-level profiling capabilities to unified attention. It works similarly to non-unified attention profiling (`VLLM_PROFILE_PROMPT` & `VLLM_PROFILE_DECODE` env vars), but now you can use `VLLM_PROFILE_UNIFIED` to pass unified buckets you want to profile - plural, because now chaining buckets works fine. You can pass `VLLM_PROFILE_UNIFIED=512,128,384,1` or something like `VLLM_PROFILE_UNIFIED="[(512,64,384,1),(512,128,384,1)]"` to profile execution of two or more buckets sequentially. I implemented this because I needed to profile memory reuse across buckets and it was super hard to do on actual benchmarks.

### DIFF
--- a/calibration/step-1-prepare-calibration-dataset.py
+++ b/calibration/step-1-prepare-calibration-dataset.py
@@ -38,6 +38,11 @@ def main(args):
             padding_side="left",
             use_fast=False,
         )
+
+        # Llama4 models in Text Only mode fail with False if use_fast=False
+        if isinstance(tokenizer, bool):
+            raise ValueError("Tokenizer is a boolean")
+
     except (OSError, ValueError, RuntimeError, ImportError):
         tokenizer = transformers.AutoTokenizer.from_pretrained(
             args.model,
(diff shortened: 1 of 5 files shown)


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -4597,6 +4613,30 @@ def warmup_multimodal_graphs(self, buckets):
                     self.graphed_buckets.add(img_arg)
                     self.log_warmup_multimodal(phase, idx, num_candidates, 1, 0, img_arg)
 
+    def _maybe_profile_unified_attn(self):
+        unified_cfg_str = os.environ.get('VLLM_PROFILE_UNIFIED', None)
```
adobrzyn: Can we add this to developer flags?

PR #815, opened 2026-01-14
Test case disablement for deepseek V2 dynamic inc

iboiko-habana:


iboiko-habana:
Fix is provided in https://github.com/vllm-project/vllm-gaudi/pull/817

PR #820, opened 2026-01-15
Dev/multi connector

yeonsily:
Enable CPU OffloadingConnector on hpu.


yeonsily:
https://github.com/vllm-project/vllm-gaudi/pull/827

PR #846, opened 2026-01-21
Add ovis models support with default buckets

testdig:


vllm_gaudi/models/ovis.py
```diff
@@ -0,0 +1,12 @@
+from vllm.config import VllmConfig
```
michalkuligowski: Copyright header missing

testdig:
I don't think my PR breaks the CI, anyone help here?

PR #850, opened 2026-01-21
Revert "skip HPU graphs for long prefills"

adobrzyn:
Reverts vllm-project/vllm-gaudi#780


yangulei:
@adobrzyn What is the motivation of this PR?

PR #856, opened 2026-01-21
Correct sliding window enabling

jbyczkow:
Currently is_interleaved just checks for at least two different layers in the model. Sliding window is used when multiple different attention layers are present in one model. For Granite 4.0-h small thats an issue since we have one type of attention layers and one type of mamba layer. Sliding window should be disabled. For Granite 4.0 it's sufficient to check if sliding_window size is defined. Proper fix would be to change is_interleaved function in the upstream. Alternativelly maybe a different function needs to be used to check for sliding window enabling.


ksmusz (approved):

michalkuligowski:
Is this needed for 0.14?

PR #868, opened 2026-01-22
[GAUDISW-243690]  Support static FP8 per-tensor scales for KV cache (without INC)

skavulya:


yiliu30:
Hi @skavulya, it looks like this PR might be a duplicate of the one where I added support for loading the Q/K/V scales for FP8 attention:
https://github.com/vllm-project/vllm-gaudi/pull/874

Previously, we supported FP8 MLA by implementing `HPUCompressedTensorsKVCacheMethodForMLA`. How about we follow the same approach here to keep things consistent?  https://github.com/vllm-project/vllm-gaudi/pull/763

skavulya:
@yiliu30 Is there a way to load static quantized models like https://huggingface.co/RedHatAI/Meta-Llama-3-8B-Instruct-FP8-KV with INC?

yiliu30:
> Hi @yiliu30 Thanks for the review. This PR is addressing a different issue which is modifying the `VLLMFP8KVCache` to support non-unit kv scales. On the other hand, `CompressedTensorsKVCacheMethod` is used loading weights. `VLLMFP8KVCache` is used when we run FP8 KV Cache without INC
> 
> https://github.com/vllm-project/vllm-gaudi/blob/d54f4c258e60904d863d9c4c540885ff443803ce/vllm_gaudi/attention/backends/hpu_attn.py#L446-L463
> 
> .

Hi @skavulya ,
The PR at https://github.com/vllm-project/vllm-gaudi/pull/874 also loads models saved as LLMC format (without using INC). It also targets non‑unit KV scales, where the K/V scale comes from the calibration process.
It feels like we’re tackling the same problem, just with different implementation details. :)

PR #877, opened 2026-01-26
Enable deepseek ocr model in gaudi plugin

HeJunyan:
Also add warmup logic for deepseek ocr model.


vllm_gaudi/models/deepseek_ocr.py
```diff
@@ -0,0 +1,171 @@
+from collections.abc import Mapping
```
michalkuligowski: Is this needed on main? Does it need a copyright header?

HeJunyan: Yes, this is needed to merge to main. But the main branch can not work for this model now. I will rebase and add copyright header then.

PR #884, opened 2026-01-27
Qwen3vl accuracy fixes

libinta:
for qwen3 vl, there is accuracy issue with multi-images within 1 request, this PR is to fix that. After fix, there are 3 paths for vision attention depending on the images count inside 1 request
1. single image, use fusedsdpa without attn mask
3.  multi-images with threshold  use fusedsdpa without attn_mask one by one
This pr also enables qwen3vl moe


afierka-intel (approved):

slokesha (approved):

michalkuligowski (approved):

wpyszka (approved):
Approved for 0.14.1

PR #885, opened 2026-01-27
Implement bucket corrector for Mamba chunk size - v0.14.1

jbyczkow:
Due to MambaMixer2 implementation requirements, all buckets used for mamba must be a multiple of mamba chunk size.


wpyszka (approved):
LGTM

PR #896, opened 2026-01-29
RowParallel NIC chunking

kamil-kaczor:
This feature enables MME call chunking in RowParallelLinear to allow the matrix multiplication to be pipelined with NIC all_reduce calls.

You can change number of chunks by using: export VLLM_ROW_PARALLEL_CHUNKS=X where 4 is the default.
You can control above what sequence len you will chunk by using: VLLM_ROW_PARALLEL_CHUNK_THRESHOLD=X where 8k is the default.


vllm_gaudi/ops/hpu_row_parallel_linear.py
```diff
@@ -0,0 +1,189 @@
+from typing import Union
```
adobrzyn: Missing license identifier

tests/unit_tests/ops/test_hpu_row_parallel_linear.py
```diff
@@ -0,0 +1,277 @@
...
+
+def _mock_config(num_chunks=1, chunk_threshold=8192):
+    """Create a mock runtime config object."""
+    cfg = MagicMock()
+    cfg.row_parallel_chunks = num_chunks
+    cfg.row_parallel_chunk_threshold = chunk_threshold
+    return cfg
+
+
+def _create_layer(num_chunks=1, chunk_threshold=8192, input_size=INPUT_SIZE, output_size=OUTPUT_SIZE, bias=False):
```
adobrzyn: Q: Don't we want to add some tests with bias as well?

vllm_gaudi/ops/hpu_row_parallel_linear.py
kamil-kaczor: fixed

PR #906, opened 2026-01-30
removed lens calculation for qwen2_5

slokesha:
Remove the lens = (cu_seqlens[1:] - cu_seqlens[:-1]).tolist() computation from the Qwen2.5 path.

This calculation is not required for Qwen2.5 and was causing a performance regression after PR #884. Removing it restores the previous performance without changing model behavior.

### DIFF
--- a/vllm_gaudi/models/qwen2_5_vl.py
+++ b/vllm_gaudi/models/qwen2_5_vl.py
@@ -83,8 +83,6 @@ class HPU_Attention:
             else:
                 return AttentionLongSequence.forward(q, k, v, mask, q_block_size, cls.softmax_mode)
         else:
-            if cu_seqlens is None:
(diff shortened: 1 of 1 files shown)


vllm_gaudi/models/qwen2_5_vl.py
```diff
@@ -77,22 +77,25 @@ def forward(cls, q, k, v, mask, cu_seqlens, qwen2_5_vl, q_block_size=64):
         Support long sequence at prompt phase
         """
         q_len = q.size(-2)
-        lens = (cu_seqlens[1:] - cu_seqlens[:-1]).tolist()
-        if mask is not None or len(lens) == 1:
-            if not qwen2_5_vl or (qwen2_5_vl and q_len < 65536):
+        if qwen2_5_vl:
```
michalkuligowski: Is this needed on main?

PR #912, opened 2026-02-02
 WA for contiguous pa with max num batched != max model len

adobrzyn:


michalkuligowski (approved):
Is this needed on main?

PR #915, opened 2026-02-03
[GAUDISW-245713] Remove bucket densification for long ctx; Edge buckets only for long ctx

kfojcik-intel:
Adding max edge buckets per query (instead of using max ctx from config) https://github.com/vllm-project/vllm-gaudi/pull/739 causes perf regression for non-long ctx cases.
Therefore, we keep this approach but only for long ctx.

### DIFF
--- a/calibration/quantization_config/maxabs_quant_dynamic_quantization_weight_pcs_pow2.json
+++ b/calibration/quantization_config/maxabs_quant_dynamic_quantization_weight_pcs_pow2.json
@@ -1,0 +1,15 @@
+{
+    "mode": "QUANTIZE",
+    "observer": "maxabs",
+    "scale_format": "CONST",
+    "scale_method": "maxabs_pcs_pow2",
+    "dynamic_quantization": true,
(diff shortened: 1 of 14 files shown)


vllm_gaudi/extension/bucketing/exponential.py
```diff
@@ -162,19 +166,20 @@ def warmup_range_with_limit(config: Tuple[int, int, int, int], long_context=Fals
...
+    #     for i in range(1, num_buckets_linear + 1):
+    #         #for i in range(1, num_buckets+1):
+    #         power_unpadded = first_step + i * tmp_step
+
+    #         if i == num_buckets and get_config().use_contiguous_pa:
+    #             bucket = bmax
+    #         else:
+    #             bucket = math.ceil(power_unpadded / bstep) * bstep
+    #         if bucket not in buckets:
+    #             buckets.add(bucket)
```
michalkuligowski: please remove commented code

iboiko-habana (approved):

PR #924, opened 2026-02-04
Add smoke test for aice branch

czhu15:
This smoke test will do a quick test on below models:
- DeepSeek-R1-Distill-Qwen-7B on one card
- Qwen3-32B on one card
- Qwen3-30B-A3B on two cards

The key idea is to make sure any changes on the PR do not break the basic features.


yupengzh-intel (approved):
lgtm

PR #926, opened 2026-02-04
Set device according to the local rank

yangulei:
### Motivation
For a typical node with 8xGaudi2E HPUs, the devices are break into two groups with 4 HPUs connected with top board each. Current random mapping between `local_rank` and `module_id` will cause HCCL failure for `world_size>4` cases.

### Changes
- Set device according to local rank.
- Use `pyhlml` to set `HABANA_VISIBLE_MODULES` to available modules. This is necessary if multiple cases with `world_size=1/2/4` wants to run on the same node simultaneously or the available `module_ids` are not start with 0.

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_worker.py
+++ b/vllm_gaudi/v1/worker/hpu_worker.py
@@ -156,8 +156,6 @@ class HPUWorker(WorkerBase):
                         selected_modules_str = ",".join(map(str, sorted(selected_modules)))
                         os.environ["HABANA_VISIBLE_MODULES"] = selected_modules_str
                         logger.warning("Using selected available modules: %s", selected_modules_str)
-        except Exception as e:
-            raise e
         finally:
             pyhlml.hlmlShutdown()
 


vllm_gaudi/v1/worker/hpu_worker.py
```diff
@@ -94,6 +94,60 @@ def __init__(
...
+                    if len(selected_modules) < self.parallel_config.world_size:
+                        raise RuntimeError(
+                            f"Not enough available modules for world_size={self.parallel_config.world_size}. "
+                            "Set HABANA_VISIBLE_MODULES to include more available modules and try again.")
+                    else:
+                        selected_modules_str = ",".join(map(str, sorted(selected_modules)))
+                        os.environ["HABANA_VISIBLE_MODULES"] = selected_modules_str
+                        logger.warning("Using selected available modules: %s", selected_modules_str)
+        except Exception as e:
+            raise e
```
czhu15: redundant exception handling here.

yangulei:
> Why there are so many changes in the examples/nixl folder? Did we tested them with the change?

The examples start multi vLLM server simultaneously. The modules for each server must be specified after this patch, or they will try to acquire the same device and results in `Device acquire failed` error.
The CI for the `main` branch used to passed with those modifications. The recent failure of CI is not relevant to this patch.

vllm_gaudi/v1/worker/hpu_worker.py
yangulei: done

PR #939, opened 2026-02-06
add context length to the high-level profile and the script to parse the profile

yangulei:


scripts/high-level_profile_to_csv.py
czhu15: suggest to rename it to high_level_profile_to_csv.py

vllm_gaudi/extension/profiler.py
```diff
@@ -61,26 +61,31 @@ def __init__(self):
...
         self.prompt_real_seq_lens.append(real_seq_lens)
+        self.prompt_real_ctx_lens.append(real_ctx_lens)
 
     def reset_prompt_seq_stats(self):
         self.prompt_real_seq_lens = []
+        self.prompt_real_ctx_lens = []
 
-    def get_counter_dict(self, cache_config, duration, seq_len, batch_size_padded, real_batch_size, prompt_batch_idx,
-                         is_prompt):
+    def get_counter_dict(self, cache_config, duration, seq_len, ctx_blocks, batch_size_padded, real_batch_size,
```
czhu15: suggest to keep old interface unchanged and add new argument to last with default value.
like
def get_counter_dict(self, cache_config, duration, seq_len, batch_size_padded, real_batch_size, prompt_batch_idx,
                         is_prompt, ctx_blocks=0)

yangulei: Like above, there are two calls for this function and both modified accordingly.

PR #942, opened 2026-02-06
cherry pick Llama4 on apply patches + QK flatten pos + perf drop

Luca-Calabria:
Cherry pick Llama4 missing fixes from #881 #862 #884 on releases/ branch


wpyszka (approved):
approved for 0.15.1

PR #975, opened 2026-02-13
Monkey-patch of Attention.forward

tzielinski-habana:
This pull request introduces a new patch for the attention layer to support custom behavior on the HPU platform, and ensures it is registered by default. Additionally, it enables a previously commented-out test in the GSM8K test suite.

The patch is needed to avoid recompilations of attention layers. These recompilations were introduced by the following PR to vllm:
https://github.com/vllm-project/vllm/pull/25954

The problem is that in the new upstream implementation, 'self.layer_name' was used in get_attention_context to retrieve the concrete attention layer instance (which was previously accessed with 'self' in Attention.forward). Because of that, every layer had to pass a different layer_name to standalone helper functions and that triggered recompilations in torch.compile. This code brings back the old access mechanism unless the following condition is met:
if self.use_output or not self.use_direct_call

For that case, more work is needed.

### DIFF
--- a/vllm_gaudi/ops/hpu_attention.py
+++ b/vllm_gaudi/ops/hpu_attention.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import torch
 
-import vllm.attention.layer as layer
+import vllm.model_executor.layers.attention.attention as layer
 from vllm.forward_context import ForwardContext, get_forward_context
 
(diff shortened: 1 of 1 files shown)


michalkuligowski (commented):
Test fail with unknow attribute:
test_init_kv_cache_with_kv_sharing_invalid_target_layer_order - KeyError: 'model_type'

iboiko-habana (approved):

PR #985, opened 2026-02-17
[WIP] Unified chunked attn shared bias improvemnts

adobrzyn:


adobrzyn:
At this point no OOM & passable but not ideal acc

PR #999, opened 2026-02-20
Fixing redundant transpose in HPUMambaMixer2

ksmusz:
We are noticing some MME transposes within the profiling, which are fixed by this change


michalkuligowski (approved):
Please cp to 0.15.1 and 0.16.0

PR #1005, opened 2026-02-23
Force CPU loading for INC quantization to prevent OOM during weight loading

kamil-kaczor:
image
Upstream changed quantization from cpu to device which caused the issue. This fixes the device it uses.

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -908,7 +908,7 @@ class HPUModelRunner(HpuKVConnectorModelRunnerMixin):
             logger.info("Bucketing is OFF.")
 
         self._PAD_SLOT_ID = -1
-        self._PAD_BLOCK_ID = 0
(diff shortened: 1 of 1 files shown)


vllm_gaudi/platform.py
```diff
@@ -150,6 +150,16 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
 
         print(f"========={compilation_config.custom_ops=}===========")
 
+        # Force CPU loading for INC quantization to prevent OOM during weight loading.
+        # INC FP8 quantization requires weights to be loaded to CPU first, then
+        # quantized and moved to device. Without this, weights are loaded directly
+        # to HPU in BF16 which causes OOM for large models.
+        model_config = vllm_config.model_config
+        is_inc_quant = (model_config is not None and model_config.quantization == "inc") or os.getenv("QUANT_CONFIG")
```
michalkuligowski: PR does not reflect the title, first commit reverted the PR in description while 2nd one reverses that

PR #1016, opened 2026-02-24
Fix HPUMambaMixer2 inheritance dependency

jbyczkow:


michalkuligowski (commented):
Is this needed for 0.15.1 also?

PR #1034, opened 2026-02-25
Enable slicing for the BF16 FusedSDPA

yangulei:
Note that this PR depends on:
- the **Boolean** attention mask introduced by https://github.com/vllm-project/vllm-gaudi/pull/1032 to get valid `m` and `linv` from the FusedSDPA kernel,
- the default query/ctx bucketing config modified in https://github.com/vllm-project/vllm-gaudi/pull/1086

### DIFF
--- a/docs/configuration/env_variables.md
+++ b/docs/configuration/env_variables.md
@@ -119,7 +119,7 @@ FusedSDPA can be split into smaller chunks to improve performance by:
 | ---------------------------------------- | -------------------------------------------------------------------------------------------- | ------------------------------------------ |
 | `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD`      | KV length threshold above which slicing is applied. Set to `-1` to disable slicing.          | `min(max_num_batched_tokens, 8192)`        |
 | `VLLM_HPU_FSDPA_SLICE_CHUNK_SIZE`        | Chunk size for `q_len` and `kv_len` in each chunk. Rounded up to the next multiple of 1024.  | `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD // 2`   |
-| `VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS` | Places each chunk in a separate graph to reduce compilation time.                            | `true`                                     |
+| `VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS` | Places each chunk in a separate graph to reduce compilation time.                            | `true` for lazy mode and `false` otherwise |
 
 !!! note
     These parameters are effective only with the linear bucketing strategy.
--- a/vllm_gaudi/extension/utils.py
+++ b/vllm_gaudi/extension/utils.py
@@ -380,11 +380,13 @@ class ModuleFusedSDPA(torch.nn.Module):
             max_ctx_pad_default = math.ceil(max_num_batched_tokens / block_size)
             max_ctx_pad = int(os.getenv("VLLM_PROMPT_CTX_BUCKET_PAD_MAX", str(max_ctx_pad_default)))
             self.num_padded_ctx_chunks = math.ceil(max_ctx_pad * block_size / self.chunk_size)
+
+            import habana_frameworks.torch as ht
+            is_lazy = ht.utils.internal.is_lazy()
             self.with_graph_breaks = os.getenv("VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS",
-                                               "true").strip().lower() in ("1", "true")
+                                               str(is_lazy)).strip().lower() in ("1", "true")
             if self.with_graph_breaks:
-                import habana_frameworks.torch as ht
-                if ht.utils.internal.is_lazy():
+                if is_lazy:
                     self.break_graph = ht.core.mark_step
                 else:
                     self.break_graph = torch._dynamo.graph_break


vllm_gaudi/extension/utils.py
```diff
@@ -166,6 +179,202 @@ def forward(
...
+            qkv_chunk_size_default = (qkv_slice_thld // 2 + 1023) // 1024 * 1024
+            qkv_chunk_size = int(os.getenv("VLLM_HPU_FSDPA_SLICE_CHUNK_SIZE", str(qkv_chunk_size_default)))
+            block_size = bucketing_manager.block_size
+            if qkv_chunk_size < block_size or qkv_chunk_size > qkv_slice_thld:
+                logger().warning(f'Invalid chunk size for FusedSDPA slicing, the chunk size should be between '
+                                 f'{block_size} and {qkv_slice_thld}, '
+                                 f'falling back to default {qkv_chunk_size_default}.')
+                qkv_chunk_size = qkv_chunk_size_default
+            if qkv_chunk_size % 1024 != 0:
+                qkv_chunk_size = (qkv_chunk_size + 1023) // 1024 * 1024
```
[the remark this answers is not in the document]
yangulei: Have to use `qkv_chunk_size = int(math.ceil(qkv_chunk_size / 1024)) * 1024` and `import math`. I prefer the integer arithmetic here.

vllm_gaudi/extension/utils.py
```diff
@@ -166,6 +179,202 @@ def forward(
...
+
+        from vllm_gaudi.extension.bucketing.linear import LinearBucketingStrategy
+        strategy = bucketing_manager.get_bucketing_strategy()
+        enable_slicing = isinstance(strategy, LinearBucketingStrategy)
+        if not enable_slicing:
+            logger().debug('Not using Linear Bucketing Strategy, slicing in FSDPA will be disabled.')
+            return False
+
+        max_num_batched_tokens = bucketing_manager.max_num_batched_tokens
+        qkv_slice_thld_default = min(max_num_batched_tokens, 8192)
```
testdig: in the case if max_num_batched_tokens is 8k or 16k, and I set VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD to 4k, then slicing cannot be enabled, is it by design?

vllm_gaudi/extension/utils.py
```diff
@@ -158,6 +160,17 @@ def forward(
         padding_side="left",
         window_size=None,
     ):
+        bs = query.shape[0]
+        q_len = query.shape[-2]
+        kv_len = key.shape[-2]
+        if (self.enable_slicing and bs == 1 and q_len != kv_len and kv_len >= self.qkv_slice_thld and is_causal
```
testdig: if slicing works only if bs == 1, shall we document it?

yangulei: `bs == 1` is always satisfied for prefills **with context**. Actually I'm not sure whether the `bs > 1` prefills with context works.

docs/configuration/env_variables.md
```diff
@@ -70,7 +70,7 @@ The following table lists the available variables with their default values:
 | Prompt | query length min (`VLLM_PROMPT_QUERY_BUCKET_MIN`)                        | `block_size`                                                                                                  |
 | Prompt | query length step (`VLLM_PROMPT_QUERY_BUCKET_STEP`)                      | `block_size`                                                                                                  |
 | Prompt | query length max (`VLLM_PROMPT_QUERY_BUCKET_MAX`)                        | `max_num_batched_tokens`                                                                                      |
-| Prompt | query length max abs padding (`VLLM_PROMPT_QUERY_BUCKET_PAD_MAX`)        | `max_num_batched_tokens`                                                                                      |
+| Prompt | query length max abs padding (`VLLM_PROMPT_QUERY_BUCKET_PAD_MAX`)        | `max_num_batched_tokens // 4`                                                                                 |
```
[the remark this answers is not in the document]
yangulei: Yes, but the `max_query_pad` and `max_ctx_pad` share the same ENVs and default values in linear bucketing, so they have to be modified simultaneously anyway.

vllm_gaudi/extension/utils.py
```diff
@@ -166,6 +179,202 @@ def forward(
...
+                    new_m = torch.maximum(last_m, chunk_m)
+                    last_linv_rescaled = (1.0 / last_linv) * torch.exp(last_m - new_m)
+                    chunk_linv_rescaled = (1.0 / chunk_linv) * torch.exp(chunk_m - new_m)
+                    last_linv = 1.0 / (last_linv_rescaled + chunk_linv_rescaled)
+                    last_out = (last_linv_rescaled * last_linv) * last_out + (chunk_linv_rescaled *
+                                                                              last_linv) * chunk_out
+                    last_m = new_m
+
+                if self.with_graph_breaks:
+                    self.break_graph()
```
czhu15: there will be 2x(num of chunks) breaks for each layer. not sure if this will introduce perf penalty when num of chunks is large...

vllm_gaudi/extension/utils.py
```diff
@@ -166,6 +179,202 @@ def forward(
...
+        enable_slicing = enable_slicing and bucketing_manager.initialized
+        if not enable_slicing:
+            logger().warning('Bucketing manager is not initialized, slicing in FSDPA will be disabled.')
+            return False
+
+        from vllm_gaudi.extension.bucketing.linear import LinearBucketingStrategy
+        strategy = bucketing_manager.get_bucketing_strategy()
+        enable_slicing = isinstance(strategy, LinearBucketingStrategy)
+        if not enable_slicing:
+            logger().debug('Not using Linear Bucketing Strategy, slicing in FSDPA will be disabled.')
```
czhu15: suggest to use info log level here. because by default debug message is not printed out.

docs/configuration/env_variables.md
```diff
@@ -85,7 +85,7 @@ The following table lists the available variables with their default values:
 | Decode | block size min (`VLLM_DECODE_BLOCK_BUCKET_MIN`)                          | `1`                                                                                                           |
 | Decode | block size step (`VLLM_DECODE_BLOCK_BUCKET_STEP`)                        | `block_size`                                                                                                  |
 | Decode | block size max (`VLLM_DECODE_BLOCK_BUCKET_MAX`)                          | `max_model_len * max_num_seqs // block_size` <br>by default or `max_blocks` <br>if `VLLM_CONTIGUOUS_PA = True`|
-| Decode | block size max abs padding (`VLLM_DECODE_BLOCK_BUCKET_PAD_MAX`)          | `max_num_batched_tokens * max_num_seqs // block_size`                                                         |
+| Decode | block size max abs padding (`VLLM_DECODE_BLOCK_BUCKET_PAD_MAX`)          | `max_num_batched_tokens * max_num_seqs // block_size // 4`                                                    |
```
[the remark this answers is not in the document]
yangulei: The logic is confusing here, `max_num_seqs // block_size` is usually 1 or 2, and the original purpose is to align `VLLM_DECODE_BLOCK_BUCKET_PAD_MAX` with `VLLM_PROMPT_QUERY_BUCKET_PAD_MAX` here.

How about change to `max_model_len * max_num_seqs // block_size // 4` to align with `VLLM_DECODE_BLOCK_BUCKET_MAX`?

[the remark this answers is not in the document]
czhu15: Yes. I also prefer to use max_model_len * max_num_seqs // block_size. it is more logic for decode case.
But not sure if // 4 or a bigger one like // 8 is better, you may decide one reasonable value.

vllm_gaudi/extension/utils.py
yangulei: We assume that the users know that slicing cannot applied for exp bucketing.
The log is in the `__init()__` of each FSDPA module, there will be `num_layers * world_size` lines of log, so I prefer to suppress the log if nothing wrong.

docs/configuration/env_variables.md
```diff
@@ -106,3 +106,20 @@ and warm-up. Recommended settings for this case are:
...
+- reducing attention-mask usage.
+
+| Parameter name                           | Description                                                                                  | Default value                              |
+| ---------------------------------------- | -------------------------------------------------------------------------------------------- | ------------------------------------------ |
+| `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD`      | KV length threshold above which slicing is applied. Set to `-1` to disable slicing.          | `min(max_num_batched_tokens, 8192)`        |
+| `VLLM_HPU_FSDPA_SLICE_CHUNK_SIZE`        | Chunk size for `q_len` and `kv_len` in each chunk. Rounded up to the next multiple of 1024.  | `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD // 2`   |
+| `VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS` | Places each chunk in a separate graph to reduce compilation time.                            | `true`                                     |
+
+!!! note
+    These parameters are effective only with the linear bucketing strategy, where the max absolute padding for `query` and `context` determines attention-mask usage.
```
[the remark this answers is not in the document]
czhu15: it is a note message. what should user do if user enables slicing with linear bucketing strategy? 
for me, "where the max absolute padding for `query` and `context` determines attention-mask usage." it is a pure internal implementation and I guess most users don't understand it :) I would suggest to remove it from the readme file. Only keep "These parameters are effective only with the linear bucketing strategy" is enough.

vllm_gaudi/extension/utils.py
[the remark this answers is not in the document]
czhu15: got it. thanks for the explaination.

vllm_gaudi/extension/utils.py
czhu15: ok. then let's keep it as debug level.

docs/configuration/env_variables.md
```diff
@@ -65,27 +65,27 @@ The following table lists the available variables with their default values:
...
 | Prompt | sequence ctx min (`VLLM_PROMPT_CTX_BUCKET_MIN`)                          | `0`                                                                                                           |
-| Prompt | sequence ctx step (`VLLM_PROMPT_CTX_BUCKET_STEP`)                        | `1`                                                                                                           |
-| Prompt | sequence ctx max (`VLLM_PROMPT_CTX_BUCKET_MAX`)                          | `(max_model_len - block_size) // block_size`                                                                  |
-| Prompt | sequence ctx max abs padding (`VLLM_PROMPT_CTX_BUCKET_PAD_MAX`)          | `max_num_batched_tokens // block_size`                                                                        |
+| Prompt | sequence ctx step (`VLLM_PROMPT_CTX_BUCKET_STEP`)                        | `2`                                                                                                           |
+| Prompt | sequence ctx max (`VLLM_PROMPT_CTX_BUCKET_MAX`)                          | `(max_model_len - block_size) / block_size`                                                                   |
```
czhu15: // should be correct, but not /

docs/configuration/env_variables.md
```diff
@@ -65,27 +65,27 @@ The following table lists the available variables with their default values:
...
+| Decode | block size min (`VLLM_DECODE_BLOCK_BUCKET_MIN`)                          | `block_size`                                                                                                  |
 | Decode | block size step (`VLLM_DECODE_BLOCK_BUCKET_STEP`)                        | `block_size`                                                                                                  |
-| Decode | block size max (`VLLM_DECODE_BLOCK_BUCKET_MAX`)                          | `max_model_len * max_num_seqs // block_size` <br>by default or `max_blocks` <br>if `VLLM_CONTIGUOUS_PA = True`|
-| Decode | block size max abs padding (`VLLM_DECODE_BLOCK_BUCKET_PAD_MAX`)          | `max_num_batched_tokens * max_num_seqs // block_size`                                                         |
+| Decode | block size max (`VLLM_DECODE_BLOCK_BCUKET_MAX`)                          | `max_model_len * max_num_seqs / block_size` <br>by default or `max_blocks` <br>if `VLLM_CONTIGUOUS_PA = True` |
+| Decode | block size max abs padding (`VLLM_DECODE_BLOCK_BUCKET_PAD_MAX`)          | `VLLM_DECODE_BLOCK_BCUKET_MAX / 4`                                                                            |
```
czhu15: should be // but no /

docs/configuration/env_variables.md
```diff
@@ -65,27 +65,27 @@ The following table lists the available variables with their default values:
 | Prompt | batch size min (`VLLM_PROMPT_BS_BUCKET_MIN`)                             | `1`                                                                                                           |
 | Prompt | batch size step (`VLLM_PROMPT_BS_BUCKET_STEP`)                           | `1`                                                                                                           |
 | Prompt | batch size max (`VLLM_PROMPT_BS_BUCKET_MAX`)                             | `max_num_prefill_seqs`                                                                                        |
-| Prompt | batch size max abs padding (`VLLM_PROMPT_BS_BUCKET_PAD_MAX`)             | `16`                                                                                                          |
+| Prompt | batch size max abs padding (`VLLM_PROMPT_BS_BUCKET_PAD_MAX`)             | `max_num_prefill_seqs / 4`                                                                                    |
```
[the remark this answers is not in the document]
yangulei: It's not for slicing. I aligned the `*BUCKET_PAD_MAX = math.ceil(BUCKET_MAX / 4)` for all the possible configurations.

PR #1047, opened 2026-02-26
Set device according to the local rank

yangulei:
Reimplement https://github.com/vllm-project/vllm-gaudi/pull/788 and fix the runtime error reported in https://github.com/vllm-project/vllm-gaudi/pull/946 by removing the range checking for module IDs.


yangulei:
@iboiko-habana
Please help to review and check if this PR fix the issue mentioned in https://github.com/vllm-project/vllm-gaudi/pull/946.

PR #1060, opened 2026-02-27
Change Qwen3VL to use HPUMMEncoderAttention

jiminha:
Cherry-pick from main: https://github.com/vllm-project/vllm-gaudi/pull/1031
Replace HPUAttention with HPUMMEncoderAttention for Qwen3-VL model.

PR #1061, opened 2026-02-27
[warmup][multimodal] Fix import error for MultiModalBudget

tvoas:
This fix is needed to avoid runtime errors when doing warmup on multimodal models. It's also present in https://github.com/vllm-project/vllm-gaudi/pull/1060 but this may not merge into main or be cherry-picked into aice branch for a while.


czhu15 (approved):
LGTM
Thanks for the quick fix.

PR #1096, opened 2026-03-05
remove range checking for module ID

yangulei:
The range checking is invalid in VMs with number of available devices less than the total number of devices on the node.


czhu15 (approved):
LGTM

PR #1097, opened 2026-03-05
fix max blocks for decode bucketing in non-contiguous pa scenario

yangulei:
## Changes:
- Fix the calculation for `max_decode_blocks` and `decode_blocks_limit`.
- Add a filter to drop the decode buckets with batched contexts larger than batched max_model_len.
- Create dummy decode warmup inputs with the limitation of `num_blocks` to avoid OOM.

### DIFF
--- a/vllm_gaudi/extension/bucketing/common.py
+++ b/vllm_gaudi/extension/bucketing/common.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import bisect
 import math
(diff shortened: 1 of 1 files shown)


vllm_gaudi/extension/bucketing/common.py
```diff
@@ -427,8 +436,8 @@ def batch_size_smaller_than_blocks(bs, query, ctx):
         },
         "decode": {
             # depends only on contiguous PA
-            True: [],
-            False: [batch_size_smaller_than_blocks],
+            True: [num_ctx_tokens_less_or_equal_batched_max_model_len],
```
czhu15: Is num_ctx_tokens_less_or_equal_batched_max_model_len required for contiguous PA too? since the title shows the fix only for non-contiguous case.

PR #1104, opened 2026-03-06
add torch inference decorator back to warmup

skaulintel:
Torch inference decorator mistakenly removed from #1054, adding it back as it was causing below issue

```
AssertionError: Optimized softmax mode is supported in recompute training mode only in causal(triangular) mask case
```


jiminha:
@jczaja  @iboiko-habana  Could you check this PR please. Without this Multimodal model warmup is failing due to the warmup running as training mode(and qkv.requires_grad set to True causing issue on the FusedSDPA).

PR #1107, opened 2026-03-06
Added Qwen CI tests

slokesha:
This PR adds a simple CI test that launches vllm serve with warmup for several Qwen models and verifies that the server starts successfully and can handle a basic request.


michalkuligowski (approved):

adobrzyn:
@slokesha - is this PR still needed? still not fixed OOM

PR #1113, opened 2026-03-07
Enable Qwen2.5 Omni Model

jjmiao1:
Test command:
image process example:
`python examples/offline_inference/vision_language.py --modality image --model-type qwen2_5_omni
`
video process example:
`python examples/offline_inference/vision_language.py --modality video --model-type qwen2_5_omni
`
audio process example:
`python examples/offline_inference/audio_language.py --model-type qwen2_5_omni
`

The audio processing currently encounters an upstream issue in vLLM and fixed in https://github.com/vllm-project/vllm/pull/35994


iboiko-habana:
@jjmiao1 please rebase code if this PR is still needed. PR will be checked ASAP after rebase

PR #1122, opened 2026-03-10
Fixes for the decode bucketing in non-contiguous pa scenario

yangulei:
## Changes:
- Fix the calculation for `max_decode_blocks` and `decode_blocks_limit`.
- Add a filter to drop the decode buckets with batched contexts larger than batched max_model_len.
- Create dummy decode warmup inputs with the limitation of `num_blocks` to avoid OOM.

### DIFF
--- a/tests/unit_tests/test_bucketing.py
+++ b/tests/unit_tests/test_bucketing.py
@@ -555,6 +555,8 @@ def test_exponential_decode_block_limit_uncapped(monkeypatch):
     assert block_cfg[2] == expected_max_decode_blocks, (
         f"Expected max_decode_blocks={expected_max_decode_blocks}, got {block_cfg[2]}")
     assert block_cfg[3] == expected_limit, (f"Expected decode_blocks_limit={expected_limit}, got {block_cfg[3]}")
+
+
 # --- Padding-aware bucketing tests ---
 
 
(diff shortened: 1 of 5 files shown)


tests/unit_tests/test_bucketing.py
```diff
@@ -189,25 +191,25 @@ def __init__(self, **kwargs):
...
+    max_num_seqs = 256
+    _, _, block_cfg = strategy.get_decode_cfgs(max_num_seqs=max_num_seqs,
                                                block_size=block_size,
                                                max_num_batched_tokens=131072,
-                                               max_model_len=91964,
+                                               max_model_len=max_model_len,
                                                max_blocks=max_blocks)
 
-    expected_max = max_blocks * 3  # 10779
+    expected_max = math.ceil(max_model_len / block_size) * max_num_seqs
```
michalkuligowski, line 212: What is the maximum for expected_max value here? This value is calculated, in fact calculation is copied from the algorithm, this test should check specific values and that filter in fact reduces buckets number

vllm_gaudi/extension/bucketing/linear.py
```diff
@@ -63,17 +63,18 @@ def get_decode_cfgs(self, max_num_seqs, block_size, max_num_batched_tokens, max_
         if contiguous_pa:
             max_decode_blocks = max_blocks
         decode_block_bucket_cfg = read_bucket_settings('decode', 'block', min=1, step=block_size, max=max_decode_blocks)
-        if decode_block_bucket_cfg[2] > max_blocks:
+        if contiguous_pa and decode_block_bucket_cfg[2] > max_blocks:
```
michalkuligowski, line 66: What is the decode bucket count now?

tests/unit_tests/test_bucketing.py
yangulei: The `expected_max` is calculated base on the corner decoding case with `context_lens = [max_model_len - 1] + [1] * (max_num_seqs - 1)`. The context lens will be padded to the max one and the total decoding blocks could be calculated with `math.ceil((max_model_len - 1) / block_size) * max_num_seqs` which could be simplified to `math.ceil(max_model_len / block_size) * max_num_seqs` as `max_model_len >> block_size` for common cases.
For an example with `max_model_len=16384`, `max_num_seqs=128` and `block_size=128`. The corner decoding case with `context_lens = [16383, 1, 1, 1, ... 1, 1, 1]` needs `ceil(16383 / 128) * 128 = ceil(16384 / 128) * 128= 16384` blocks after padding.

vllm_gaudi/extension/bucketing/linear.py
yangulei: This is a bug fix for the missing buckets which cause "not warmed-up" warnings for non-contiguous PA cases. The actual number of decode buckets for linear bucketing is sensitive to the `*_BUCKET_STEP_*` configuration. And the default settings usually produce too many buckets that needs hours even days to warmup for cases with long `max_model_len`.

tests/unit_tests/test_bucketing.py
yangulei: And the UT for the newly added filter is added. Thanks for remind that.

yangulei:
@adobrzyn @kamil-kaczor
Could you help to review this PR? This could solve the issue GAUDISW-247226.

kamil-kaczor (approved):
lgtm

PR #1126, opened 2026-03-10
Fix multimodal prefill batching for 2D padded inputs

afierka-intel:
Enable batching of multimodal prefill requests in Plugin v1 by fixing the position calculation in _gather_mm_embeddings for 2D padded batches.

Previously, when max_prefill_batch_size > 1 and multiple multimodal requests were batched together in a 2D padded tensor [bs, padded_seq], the is_mm_embed positions were computed using contiguous offsets (req_start_idx += num_scheduled_tokens). This is incorrect for 2D padded layout where each request occupies a full padded row.

Changes:
- Add padded_seq_len parameter to _gather_mm_embeddings to compute positions as batch_row * padded_seq_len + offset for 2D batches
- Update _get_model_mm_inputs to detect 2D batched inputs (bs > 1) and pass the correct total_num_scheduled_tokens and padded_seq_len
- Update comments in _execute_mm_encoder to reflect that encoder outputs are cached by mm_hash, so ordering does not need to match the request order in the prefill batch

This enables proper multimodal prefill batching, improving throughput for workloads with many small images (e.g. MinerU PDF processing with qwen2-vl) where the previous one-at-a-time processing was 58% slower.


adobrzyn:
**Finding #2** 🟢 Nit

**File:** `vllm_gaudi/v1/worker/hpu_model_runner.py` (Line ~1580, unchanged line in loop)

**Issue:** When `padded_seq_len is not None`, position is calculated via `batch_row * padded_seq_len` instead of `req_start_idx`. However, `req_start_idx += num_scheduled_tokens` still executes unconditionally at the end of each loop iteration — making it dead code in the 2D padded path.

**Suggested fix:**
Either guard the accumulation or add a brief comment clarifying it's only relevant for the non-padded path:

```python
            if padded_seq_len is None:
                req_start_idx += num_scheduled_tokens
```

This is purely a readability nit — no functional impact.

---

PR #1144, opened 2026-03-12
Fix multimodal warmup and add multimodal profiling

jiminha:
Fix multimodal warmup and profiling

    - Fix multimodal warmup to use specified image dimensions (width/height)
      instead of default parameters.   
      It seems https://github.com/vllm-project/vllm-gaudi/pull/822 was removed mm_option, which has width/height, so the multimodal warmup was done with default dimensions regardless of the bucket calculated.  
    - Create multimodal features when limit_per_prompt is configured,
      and it's prompt profile mode.
    - Add unique identifiers to multimodal embeddings to prevent caching
      across the 3 profiling warmup runs


jinyouzhi (approved):

jiminha:
Confirmed with @jinyouzhi  that this Qwen2.5 crash issue is not related to this change and seen in baseline.

jiminha:
I'm closing this since main part of the multimodal warmup fix is integrated as part of https://github.com/vllm-project/vllm-gaudi/pull/1392

PR #1154, opened 2026-03-13
Add the option to use Boolean attention mask

yangulei:


yangulei:
Close as no obvious perf gain and have accuracy issue when using with INC fp8.

PR #1155, opened 2026-03-13
Enable slicing for the FusedSDPA

yangulei:
FusedSDPA can be split into smaller chunks to improve performance while using the padding-aware bucketing strategy which guarantees the max absolute padding in the sequence and context dimensions.

## Usage
| Parameter name                           | Description                                                                                  | Default value                               |
| ---------------------------------------- | -------------------------------------------------------------------------------------------- | ------------------------------------------- |
| `VLLM_HPU_FSDPA_SLICE_ENABLED`           | Enable the slicing.                                                                          | `True` for padding-aware bucketing strategy |
| `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD`      | KV length threshold above which slicing is applied.                                          | `min(max_num_batched_tokens, 8192)`         |
| `VLLM_HPU_FSDPA_SLICE_CHUNK_SIZE`        | Chunk size for `q_len` and `kv_len` in each chunk. Rounded up to the next multiple of 1024.  | `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD // 2`    |
| `VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS` | Places each chunk in a separate graph to reduce compilation time.                            | `true` for lazy mode and `false` otherwise  |

> [!IMPORTANT]
> These parameters are effective only with the padding-aware bucketing strategy set by `VLLM_BUCKETING_STRATEGY="pad"`.

## Implementation
Take the prefix-prefill with `[bs, query, context] = [1, 9037, 8832]` as an example. The prefill shape will first be padded to `[1, 10880, 11008]` by the bucketing. The attention mask will be looks like:
fsdpa_apc_b1_s9037_c8832_pb1_ps10880_pc11008_attention_mask

> Not that there are padding in query and context dimensions.

The original implementation pass the full attention mask to the FusedSDPA kernel.

This PR introduced an implementation to calculate the FSDPA in chunks by slicing the `Q`, `K` and `V` as below:
fsdpa_apc_b1_s9037_c8832_pb1_ps10880_pc11008_bf16_SliceQKV

Where the color of the rectangles differentiate the `is_causal` and `attn_mask` parameters passed to the FusedSDPA kernel:
- `rgb(255,0,0)`: `is_causal=False` and `attn_mask is not None`
- `rgb(255,255,0)`: `is_causal=True` and `attn_mask=None`
- `rgb(255,0,255)`: `is_causal=False` and `attn_mask=None`

In this way, most of the chunks call the FusedSDPA without attention mask to get better performance, and the graph for the chunks might be reused across different buckets to reduce the warmup duration.

## Dependencies
- https://github.com/vllm-project/vllm-gaudi/pull/762 as the number of chunks with padding is determined by the `PAD_MAX` for the query and context.

---
### Thanks @Wei-Lin-Intel for the original idea and the detailed behavior of the FusedSDPA kernel.

### DIFF
--- a/docs/configuration/env_variables.md
+++ b/docs/configuration/env_variables.md
@@ -123,12 +123,12 @@ and warm-up. Recommended settings for this case are:
 
 FusedSDPA can be split into smaller chunks to improve performance while using the padding-aware bucketing strategy which guarantees the max absolute padding in the sequence and context dimensions.
 
-| Parameter name                           | Description                                                                                  | Default value                              |
-| ---------------------------------------- | -------------------------------------------------------------------------------------------- | ------------------------------------------ |
+| Parameter name                           | Description                                                                                  | Default value                               |
+| ---------------------------------------- | -------------------------------------------------------------------------------------------- | ------------------------------------------- |
 | `VLLM_HPU_FSDPA_SLICE_ENABLED`           | Enable the slicing.                                                                          | `True` for padding-aware bucketing strategy |
-| `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD`      | KV length threshold above which slicing is applied.                                          | `min(max_num_batched_tokens, 8192)`        |
-| `VLLM_HPU_FSDPA_SLICE_CHUNK_SIZE`        | Chunk size for `q_len` and `kv_len` in each chunk. Rounded up to the next multiple of 1024.  | `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD // 2`   |
-| `VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS` | Places each chunk in a separate graph to reduce compilation time.                            | `true` for lazy mode and `false` otherwise |
+| `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD`      | KV length threshold above which slicing is applied.                                          | `min(max_num_batched_tokens, 8192)`         |
+| `VLLM_HPU_FSDPA_SLICE_CHUNK_SIZE`        | Chunk size for `q_len` and `kv_len` in each chunk. Rounded up to the next multiple of 1024.  | `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD // 2`    |
+| `VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS` | Places each chunk in a separate graph to reduce compilation time.                            | `true` for lazy mode and `false` otherwise  |
 
 !!! note
     These parameters are effective only with the padding-aware bucketing strategy set by `VLLM_BUCKETING_STRATEGY="pad"`.
--- a/docs/configuration/env_variables.md
+++ b/docs/configuration/env_variables.md
@@ -132,3 +132,10 @@ FusedSDPA can be split into smaller chunks to improve performance while using th
 
 !!! note
     These parameters are effective only with the padding-aware bucketing strategy set by `VLLM_BUCKETING_STRATEGY="pad"`.
+
+The slicing only activated if all the following additional conditions are satisfied:
+- The batch size should be 1.
+- The query length and KV length should be different, i.e. the normal causal prefill will route to the default dispatch for better performance.
+- It's causal attention model.
+- The padding side is 'right'.
+- no sliding window nor sinks.
--- a/docs/configuration/env_variables.md
+++ b/docs/configuration/env_variables.md
@@ -118,3 +118,17 @@ and warm-up. Recommended settings for this case are:
 
 !!! note
     If the model config specifies a high `max_model_len`, set it to the sum of `input_tokens` and `output_tokens`, rounded up to a multiple of `block_size` according to actual requirements.
+
+## Additional Performance Tuning Parameters for the FusedSDPA Kernel with Padding-Aware Bucketing
+
+FusedSDPA can be split into smaller chunks to improve performance while using the padding-aware bucketing strategy which guarantees the max absolute padding in the sequence and context dimensions.
+
+| Parameter name                           | Description                                                                                  | Default value                              |
+| ---------------------------------------- | -------------------------------------------------------------------------------------------- | ------------------------------------------ |
+| `VLLM_HPU_FSDPA_SLICE_ENABLED`           | Enable the slicing.                                                                          | `True` for padding-aware bucketing strategy |
+| `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD`      | KV length threshold above which slicing is applied.                                          | `min(max_num_batched_tokens, 8192)`        |
+| `VLLM_HPU_FSDPA_SLICE_CHUNK_SIZE`        | Chunk size for `q_len` and `kv_len` in each chunk. Rounded up to the next multiple of 1024.  | `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD // 2`   |
+| `VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS` | Places each chunk in a separate graph to reduce compilation time.                            | `true` for lazy mode and `false` otherwise |
+
+!!! note
+    These parameters are effective only with the padding-aware bucketing strategy set by `VLLM_BUCKETING_STRATEGY="pad"`.
--- a/docs/configuration/env_variables.md
+++ b/docs/configuration/env_variables.md
@@ -125,7 +125,7 @@ FusedSDPA can be split into smaller chunks to improve performance while using th
 
 | Parameter name                           | Description                                                                                  | Default value                               |
 | ---------------------------------------- | -------------------------------------------------------------------------------------------- | ------------------------------------------- |
-| `VLLM_HPU_FSDPA_SLICE_ENABLED`           | Enable the slicing.                                                                          | `True` for padding-aware bucketing strategy |
+| `VLLM_HPU_FSDPA_SLICE_ENABLED`           | Enable the slicing.                                                                          | `True` when using padding-aware bucketing strategy with bucketing enabled, merged prefill disabled, and FusedSDPA kernel available |
 | `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD`      | KV length threshold above which slicing is applied.                                          | `min(max_num_batched_tokens, 8192)`         |
 | `VLLM_HPU_FSDPA_SLICE_CHUNK_SIZE`        | Chunk size for `q_len` and `kv_len` in each chunk. Rounded up to the next multiple of 1024.  | `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD // 2`    |
 | `VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS` | Places each chunk in a separate graph to reduce compilation time.                            | `true` for lazy mode and `false` otherwise  |
@@ -133,9 +133,9 @@ FusedSDPA can be split into smaller chunks to improve performance while using th
 !!! note
     These parameters are effective only with the padding-aware bucketing strategy set by `VLLM_BUCKETING_STRATEGY="pad"`.
 
-The slicing only activated if all the following additional conditions are satisfied:
+The slicing is only activated if all the following additional conditions are satisfied:
 - The batch size should be 1.
 - The query length and KV length should be different, i.e. the normal causal prefill will route to the default dispatch for better performance.
-- It's causal attention model.
+- It's a causal attention model.
 - The padding side is 'right'.
-- no sliding window nor sinks.
+- No sliding window nor sinks (BF16 only; FP8 does not support sinks).
--- a/tests/unit_tests/test_fsdpa_slicing.py
+++ b/tests/unit_tests/test_fsdpa_slicing.py
@@ -415,6 +415,147 @@ class TestSetupSlicing:
         assert result is True
         assert base._with_graph_breaks is True
 
+    @patch('vllm_gaudi.extension.utils.get_config')
+    @patch('vllm_gaudi.extension.bucketing.common.HPUBucketingManager.get_instance')
+    def test_disabled_when_bucketing_off(self, mock_get_instance, mock_get_config):
+        mock_get_config.return_value = _make_config(use_bucketing=False)
+        mock_get_instance.return_value = None
+
+        from vllm_gaudi.extension.utils import SlicedFusedSDPABase
+        base = SlicedFusedSDPABase.__new__(SlicedFusedSDPABase)
+        result = base._setup_slicing()
+        assert result is False
+
+    @patch('habana_frameworks.torch.utils.internal.is_lazy', return_value=False)
+    @patch('vllm_gaudi.extension.utils.get_config')
+    @patch('vllm_gaudi.extension.bucketing.common.HPUBucketingManager.get_instance')
+    def test_threshold_lte_block_size_raises(self, mock_get_instance, mock_get_config, mock_is_lazy, monkeypatch):
+        """slice_thld must be greater than block_size."""
+        monkeypatch.setenv("VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD", "128")
+        mock_get_config.return_value = _make_config()
+        mock_get_instance.return_value = _MockBucketingManager(max_num_batched_tokens=8192, block_size=128)
+
+        from vllm_gaudi.extension.utils import SlicedFusedSDPABase
+        base = SlicedFusedSDPABase.__new__(SlicedFusedSDPABase)
+        with pytest.raises(AssertionError):
+            base._setup_slicing()
+
+    @patch('habana_frameworks.torch.utils.internal.is_lazy', return_value=False)
+    @patch('vllm_gaudi.extension.utils.get_config')
+    @patch('vllm_gaudi.extension.bucketing.common.HPUBucketingManager.get_instance')
+    def test_threshold_lte_1024_raises(self, mock_get_instance, mock_get_config, mock_is_lazy, monkeypatch):
+        """slice_thld must be greater than 1024."""
+        monkeypatch.setenv("VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD", "1024")
+        mock_get_config.return_value = _make_config()
+        mock_get_instance.return_value = _MockBucketingManager(max_num_batched_tokens=8192, block_size=128)
+
+        from vllm_gaudi.extension.utils import SlicedFusedSDPABase
+        base = SlicedFusedSDPABase.__new__(SlicedFusedSDPABase)
+        with pytest.raises(AssertionError):
+            base._setup_slicing()
+
+
+# ---------------------------------------------------------------------------
+# Manual module construction tests (simulates production _setup_slicing result)
+# ---------------------------------------------------------------------------
+
+
+def _make_sliced_bf16(chunk_size, num_padded_query_chunks=0, num_padded_ctx_chunks=0, with_graph_breaks=False):
+    """Build a SlicedFusedSDPA bypassing _setup_slicing (for testing / HPU accuracy)."""
+    from vllm_gaudi.extension.utils import SlicedFusedSDPA
+    with patch('vllm_gaudi.extension.utils.SlicedFusedSDPABase._setup_slicing', return_value=True):
+        module = SlicedFusedSDPA()
+    module.enable_slicing = True
+    module.chunk_size = chunk_size
+    module.slice_thld = 0
+    module.num_padded_query_chunks = num_padded_query_chunks
+    module.num_padded_ctx_chunks = num_padded_ctx_chunks
+    module._with_graph_breaks = with_graph_breaks
+    if with_graph_breaks:
+        import habana_frameworks.torch as ht
+        if ht.utils.internal.is_lazy():
+            module._break_graph = ht.core.mark_step
+        else:
+            module._break_graph = torch._dynamo.graph_break
+    return module
+
+
+def _make_sliced_fp8(chunk_size,
+                     num_padded_query_chunks=0,
+                     num_padded_ctx_chunks=0,
+                     *,
+                     d_scale_q,
+                     d_scale_k,
+                     d_scale_v,
+                     d_scale_output,
+                     scale_amax,
+                     descale_amax,
+                     with_graph_breaks=False):
+    """Build a SlicedFP8FusedSDPA bypassing _setup_slicing (for testing / HPU accuracy)."""
+    from vllm_gaudi.extension.utils import SlicedFP8FusedSDPA
+
+    # Create a lightweight parent-like namespace to hold scale tensors
+    class _ScaleHolder:
+        pass
+
+    parent = _ScaleHolder()
+    parent.d_scale_q = d_scale_q
+    parent.d_scale_k = d_scale_k
+    parent.d_scale_v = d_scale_v
+    parent.d_scale_output = d_scale_output
+    parent.scale_amax = scale_amax
+    parent.descale_amax = descale_amax
+
+    with patch('vllm_gaudi.extension.utils.SlicedFusedSDPABase._setup_slicing', return_value=True):
+        module = SlicedFP8FusedSDPA(parent)
+    module.enable_slicing = True
+    module.chunk_size = chunk_size
+    module.slice_thld = 0
+    module.num_padded_query_chunks = num_padded_query_chunks
+    module.num_padded_ctx_chunks = num_padded_ctx_chunks
+    module._with_graph_breaks = with_graph_breaks
+    if with_graph_breaks:
+        import habana_frameworks.torch as ht
+        if ht.utils.internal.is_lazy():
+            module._break_graph = ht.core.mark_step
+        else:
+            module._break_graph = torch._dynamo.graph_break
+    return module
+
+
+class TestManualModuleConstruction:
+    """Tests for building sliced modules with manual attribute setup (bypassing _setup_slicing)."""
+
+    def test_bf16_manual_init_sets_attributes(self):
+        module = _make_sliced_bf16(chunk_size=4096,
+                                   num_padded_query_chunks=1,
+                                   num_padded_ctx_chunks=2,
+                                   with_graph_breaks=False)
+        assert module.enable_slicing is True
+        assert module.chunk_size == 4096
+        assert module.num_padded_query_chunks == 1
+        assert module.num_padded_ctx_chunks == 2
+        assert module._with_graph_breaks is False
+
+    def test_fp8_manual_init_sets_scales(self):
+        scales = {
+            'd_scale_q': torch.tensor(1.0),
+            'd_scale_k': torch.tensor(1.0),
+            'd_scale_v': torch.tensor(1.0),
+            'd_scale_output': torch.tensor(1.0),
+            'scale_amax': torch.tensor(1.0),
+            'descale_amax': torch.tensor(1.0),
+        }
+        module = _make_sliced_fp8(chunk_size=4096, num_padded_query_chunks=1, num_padded_ctx_chunks=2, **scales)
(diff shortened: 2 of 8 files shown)


yangulei:
@afierka-intel 
This PR is re-implemented to enable slicing for both bf16 and fp8 FusedSDPA. Please help to review.

cc: @czhu15

vllm_gaudi/extension/utils.py
```diff
@@ -149,12 +151,237 @@ def forward(self, input, other, **kwargs):
...
+            return False
+
+        if get_config().merged_prefill:
+            logger().warning_once(
+                'FusedSDPA slicing is not compatible with merged prefill, slicing in FusedSDPA will be disabled.')
+            return False
+
+        from vllm_gaudi.extension.bucketing.common import get_bucketing_manager
+        bucketing_manager = get_bucketing_manager()
+        assert bucketing_manager is not None and bucketing_manager.initialized, 'Bucketing manager should be instantiated and initialized to enable FusedSDPA slicing.'
```
michalkuligowski, line 189: please fix the logic for env with respect to pad bucketing If a user enables VLLM_BUCKETING_STRATEGY=pad but disables bucketing (VLLM_ENABLE_BUCKETING=0), the feature flag enable_fsdpa_slicing evaluates to True but the bucketing manager won't exist, causing a hard crash at model init

vllm_gaudi/extension/utils.py
```diff
@@ -149,12 +151,237 @@ def forward(self, input, other, **kwargs):
...
+                f'The FusedSDPA slice sequence length threshold {slice_thld} is less than the default {slice_thld_default} which is not recommended.'
+            )
+
+        # default to half of the threshold and round up by 1024
+        chunk_size_default = math.ceil(slice_thld // 2 / 1024) * 1024
+        chunk_size = int(os.getenv("VLLM_HPU_FSDPA_SLICE_CHUNK_SIZE", str(chunk_size_default)))
+        assert chunk_size > block_size and chunk_size <= slice_thld, 'Invalid FusedSDPA slice chunk size, the chunk size should be between the block size and the slice sequence length threshold.'
+        if chunk_size % 1024 != 0:
+            chunk_size = math.ceil(chunk_size / 1024) * 1024
+            logger().warning_once('Rounded up the chunk size for FusedSDPA slicing to the next multiple of 1024.')
```
michalkuligowski, line 213: When max_num_batched_tokens is small (e.g., 512), slice_thld=512, chunk_size_default = ceil(256/1024)*1024 = 1024 > 512, which fails the assertion chunk_size <= slice_thld

vllm_gaudi/extension/features.py
```diff
@@ -117,5 +123,9 @@ def get_features():
               All(VersionRange(">=1.24.0.460"), MinPackageVersion("neural_compressor_pt", "3.7")),
               env_var_type=boolean),
         Value('use_hpu_aligned_scale', False, env_var='HPU_ALIGNED_SCALE', env_var_type=boolean),
+        Value('enable_fsdpa_slicing',
+              All(Eq('bucketing_strategy', 'pad'), Disabled('merged_prefill'), Kernel(fsdpa)),
```
michalkuligowski: this should use Eq('use_bucketing', True)

vllm_gaudi/extension/utils.py
```diff
@@ -182,7 +427,181 @@ def forward(
                                                 (-1, -1), sinks)
 
 
-class ModuleFP8FusedSDPA(torch.nn.Module):
+class SlicedFP8FusedSDPA(torch.nn.Module):
```
michalkuligowski: SlicedFusedSDPA.forward() and SlicedFP8FusedSDPA.forward() share ~80% identical logic (chunk iteration, online softmax rescaling, graph breaks). The only differences are: (a) the kernel call, (b) FP8 dequantization of chunk outputs, and (c) linv scaling factor (128.0 for fast softmax mode in FP8). Consider extracting the shared chunk iteration/merge pattern into a helper or base class method, with the kernel call as a customization point.

vllm_gaudi/extension/utils.py
```diff
@@ -198,13 +617,28 @@ def __init__(self, fusedSDPA):
...
         self.d_scale_k = torch.tensor(1.0, dtype=torch.float32)
         self.d_scale_v = torch.tensor(1.0, dtype=torch.float32)
+        self.d_scale_output = torch.tensor(1.0, dtype=torch.float32)
+        if self.enable_slicing:
+            self._sliced_module = SlicedFP8FusedSDPA(self.chunk_size, self.num_padded_query_chunks,
+                                                     self.num_padded_ctx_chunks, self.d_scale_q, self.d_scale_k,
+                                                     self.d_scale_v, self.d_scale_output, self.scale_amax,
+                                                     self.descale_amax, self._with_graph_breaks)
+
+    def _sync_sliced_module_scales(self) -> None:
```
michalkuligowski: The _sync_sliced_module_scales() copies references, but any future code that reassigns self.d_scale_q (rather than mutating in-place) after the sync would leave the sliced module with stale scales. Consider:

Using @property accessors on the sliced module that read from the parent, or
At minimum, adding a comment near ModuleFP8FusedSDPA.__init__ warning that scale reassignment must be followed by _sync_sliced_module_scales()

docs/configuration/env_variables.md
```diff
@@ -114,3 +114,17 @@ and warm-up. Recommended settings for this case are:
...
 !!! note
     If the model config specifies a high `max_model_len`, set it to the sum of `input_tokens` and `output_tokens`, rounded up to a multiple of `block_size` according to actual requirements.
+
+## Additional Performance Tuning Parameters for the FusedSDPA Kernel with Padding-Aware Bucketing
+
+FusedSDPA can be split into smaller chunks to improve performance while using the padding-aware bucketing strategy which guarantees the max absolute padding in the sequence and context dimensions.
+
+| Parameter name                           | Description                                                                                  | Default value                               |
+| ---------------------------------------- | -------------------------------------------------------------------------------------------- | ------------------------------------------- |
+| `VLLM_HPU_FSDPA_SLICE_ENABLED`           | Enable the slicing.                                                                          | `True` for padding-aware bucketing strategy |
```
michalkuligowski: The documentation implies slicing activates whenever KV length exceeds the threshold, but the actual conditions are stricter: bs==1, q_len != kv_len, causal+mask, right padding, no sliding window, no sinks. Users tuning these parameters should know when slicing actually fires.

vllm_gaudi/extension/utils.py
yangulei: Done, thanks!

vllm_gaudi/extension/utils.py
yangulei: Done by adding `assert slice_thld > 1024` before the calculation of `chunk_size`.

vllm_gaudi/extension/features.py
yangulei: Done, thanks~

vllm_gaudi/extension/utils.py
yangulei: Done, thanks!

vllm_gaudi/extension/utils.py
yangulei: Done, thanks!

docs/configuration/env_variables.md
yangulei: Updated, thanks~

kamil-kaczor:
**Finding 1** 🔴 **Critical** · `utils.py — _setup_slicing`

The documentation states the default for `VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS` is "`true` for lazy mode and `false` otherwise", but the code simply reads:

```python
self._with_graph_breaks = get_config().VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS
```

There is no lazy-mode-aware default logic. If the env var is not set, the `Env('VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS', boolean)` flag likely resolves to `None`/`False`, meaning graph breaks are **always** disabled by default — contradicting the docs. The test `test_graph_breaks_enabled_in_lazy_mode` expects `_with_graph_breaks is True` without setting the env var, which suggests this test might fail, or there's hidden Config resolution logic not visible in this diff.

**Suggestion:** Add explicit lazy-mode default:
```python
import habana_frameworks.torch.utils.internal as ht_internal
default_with_gb = ht_internal.is_lazy()
self._with_graph_breaks = get_config().VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS
if self._with_graph_breaks is None:
    self._with_graph_breaks = default_with_gb
```

---

kamil-kaczor:
**Finding 2** 🔴 **Critical** · `ops.py:L415-L418` → `utils.py:L399-L402`

The `is_causal = False` / `valid_seq_lengths = None` workaround has been removed from `_fsdpa_prompt_attention` (in `ops.py`) and relocated to both `ModuleFusedSDPA.forward` and `ModuleFP8FusedSDPA.forward` (in `utils.py`). Critically, the **sliced path bypasses this workaround** and passes `is_causal=True` with mask chunks directly to the kernel.

This is a *semantic* change — previously `_fsdpa_prompt_attention` always disabled causal when a mask was present; now only the non-sliced path does so. If any other caller of `_fsdpa_prompt_attention` relied on this workaround being applied at the ops layer (not the module layer), it could break silently.

**Suggestion:** Add a comment in `ops.py` explaining why the workaround was moved, and verify no other callers depend on it.

---

kamil-kaczor:
**Finding 3** 🟡 **Medium** · `utils.py — SlicedFP8FusedSDPA.chunk_kernel`

```python
linv = linv.to(torch.float32) * (128.0 if sm == "fast" else 1.0)
```

The factor `128.0` when `softmax_mode == "fast"` is unexplained. Is this compensating for an FP8 scaling factor in the `fp8_sdpa_recomp_fwd` kernel? If the kernel's fast-mode softmax applies internal scaling, this is a coupling that will silently break if the kernel changes.

**Suggestion:** Add a comment explaining the origin of `128.0`, e.g., `# fast softmax mode returns linv scaled by 1/128; rescale to true linv`.

---

kamil-kaczor:
**Finding 4** 🟡 **Medium** · `utils.py — _merge_chunk`

```python
last_linv_rescaled = (1.0 / last_linv) * torch.exp(last_m - new_m)
chunk_linv_rescaled = (1.0 / chunk_linv) * torch.exp(chunk_m - new_m)
```

Division by `last_linv`/`chunk_linv` will produce `inf`/`nan` if either value is zero. While in practice attention weights should prevent this, an all-masked chunk (e.g., from excessive padding) could produce `linv = 0`. Consider either a softmax stability guard (`linv = torch.clamp(linv, min=1e-8)`) or an assert/early check.

---

kamil-kaczor:
**Finding 5** 🟡 **Medium** · `utils.py — ModuleFP8FusedSDPA.forward`

```python
-        qinput = self.quant_input(query, self.scale_q)
+        qinput = self.quant_input(query, self.scale_q).detach()
```

Three `.detach()` calls added to `ModuleFP8FusedSDPA.forward` break the autograd graph. While vLLM is inference-only, this is a behavioral change from the current code. Why was this added? Was it needed for the sliced path, or to fix a gradient-tracking memory issue? Should be documented.

---

kamil-kaczor:
**Finding 6** 🟡 **Medium** · `utils.py — dequant_output removed`

`ModuleFP8FusedSDPA.dequant_output(self, output, scale)` is removed as a public method. The new `SlicedFP8FusedSDPA._dequant_output` is private and uses a different signature (no `scale` param — reads from `self._parent.d_scale_output`).

If any external code calls `module.dequant_output()`, this is a breaking change. If confirmed unused externally, note removal in the PR description.

---

kamil-kaczor:
**Finding 7** 🟡 **Medium** · `test_fsdpa_slicing.py` — accuracy thresholds

- BF16: cosine similarity > **0.92**, max abs diff < **0.45**
- FP8: cosine similarity > **0.74**, max abs diff < **0.45**

A cosine similarity of 0.74 means ~25% divergence from ground truth. While FP8 quantization introduces noise, these thresholds are very permissive. Were they derived empirically?

**Suggestion:** Add comments with observed baseline values, e.g.:
```python
# Baseline observed: cos_sim ~0.82, max_diff ~0.35
# Threshold set with 10% margin
assert cos_sim > 0.74
```

---

kamil-kaczor:
**Finding 8** 🟡 **Medium** · `Qwen3-30B-A3B-FP8-Static-longbench.yaml`

```yaml
metrics:
  name: qa_f1_score,none
  value: 0.12
```

A QA F1 score of 12% is extremely low. Is this the expected baseline for this model on longbench_qasper with `limit: 20`? If so, documentation explaining why should be added. If not, this may indicate a configuration issue (e.g., the model needs think-mode disabled, or different generation parameters).

---

kamil-kaczor:
**Finding 9** 🟢 **Low** · `utils.py:L399-L406`, `utils.py:L540-L547`

```python
if (self._sliced_module.enable_slicing and kv_len >= self._sliced_module.slice_thld \
        and bs == 1 \
```

PEP 8 recommends implicit line continuation inside parentheses rather than `\`. The parentheses are already present, making the backslashes redundant.

---

kamil-kaczor:
**Finding 10** 🟢 **Low** · `features.py`

`VLLM_HPU_FSDPA_SLICE_ENABLED` is registered both in `get_user_flags()` (as an `Env` flag) and as the `env_var` on the `enable_fsdpa_slicing` Value in `get_features()`. Verify this doesn't cause double-resolution or config conflicts in the Config system.

---

kamil-kaczor:
**Finding 11** 🟢 **Low** · `test_fsdpa_slicing.py` (1411 lines)

The test file covers feature flags, setup, dispatch, accuracy, and graph breaks — 5 distinct areas. Consider splitting into smaller focused files for maintainability (e.g., `test_fsdpa_slicing_config.py`, `test_fsdpa_slicing_accuracy.py`, `test_fsdpa_slicing_graph_breaks.py`).

---

docs/configuration/env_variables.md
```diff
@@ -115,3 +115,24 @@ and warm-up. Recommended settings for this case are:
...
+| `VLLM_HPU_FSDPA_SLICE_WITH_GRAPH_BREAKS` | Places each chunk in a separate graph to reduce compilation time.                            | `true` for lazy mode and `false` otherwise  |
+
+!!! note
+    These parameters are effective only with the padding-aware bucketing strategy set by `VLLM_BUCKETING_STRATEGY="pad"`.
+
+The slicing is only activated if all the following additional conditions are satisfied:
+- The batch size should be 1.
+- The query length and KV length should be different, i.e. the normal causal prefill will route to the default dispatch for better performance.
+- It's a causal attention model.
+- The padding side is 'right'.
```
kamil-kaczor, line 140: why only right padding?

tests/unit_tests/test_fsdpa_slicing.py
```diff
@@ -0,0 +1,1404 @@
+# SPDX-License-Identifier: Apache-2.0
+###############################################################################
+# Copyright (C) 2025 Intel Corporation
```
kamil-kaczor: 2026

tests/unit_tests/test_fsdpa_slicing.py
```diff
@@ -0,0 +1,1404 @@
...
+class TestSetupSlicing:
+    """Tests for SlicedFusedSDPABase._setup_slicing method."""
+
+    @patch('vllm_gaudi.extension.utils.get_config')
+    @patch('vllm_gaudi.extension.bucketing.common.HPUBucketingManager.get_instance')
+    def test_disabled_when_feature_flag_off(self, mock_get_instance, mock_get_config):
+        mock_get_config.return_value = _make_config(enable_fsdpa_slicing=False)
+        mock_get_instance.return_value = None
+
+        from vllm_gaudi.extension.utils import SlicedFusedSDPABase
```
kamil-kaczor: why do we keep repeating the import in each test?

kamil-kaczor: Maybe add it as a function call or something because it bloats up the file.

tests/unit_tests/test_fsdpa_slicing.py
```diff
@@ -0,0 +1,1404 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Manual module construction tests (simulates production _setup_slicing result)
+# ---------------------------------------------------------------------------
+
+
+def _make_sliced_bf16(chunk_size, num_padded_query_chunks=0, num_padded_ctx_chunks=0, with_graph_breaks=False):
+    """Build a SlicedFusedSDPA bypassing _setup_slicing (for testing / HPU accuracy)."""
+    from vllm_gaudi.extension.utils import SlicedFusedSDPA
```
kamil-kaczor: in general, is there a benefit from importing inside the tests vs at the top? Is it reinitializing the FusedSDPA or is there some other need?

tests/unit_tests/test_fsdpa_slicing.py
```diff
@@ -0,0 +1,1404 @@
...
+        k = torch.randn(1, 4, 8192, 64)  # kv_len=8192 >= threshold
+        v = torch.randn(1, 4, 8192, 64)
+        mask = torch.zeros(1, 1, 2048, 8192)
+
+        module._sliced_module.return_value = torch.zeros(1, 4, 2048, 64)
+        module.forward(q, k, v, mask, 0.0, True, None, 'fast', True, None, padding_side='right')
+        module._sliced_module.assert_called_once()
+
+    def test_default_path_when_slicing_disabled(self):
+        import torch
```
kamil-kaczor: Probably the import is also not needed in every fun call.

vllm_gaudi/extension/utils.py
```diff
@@ -149,12 +151,246 @@ def forward(self, input, other, **kwargs):
...
+            chunk_kernel_fn: Callable
+                ``(q, k, v, mask, dropout_p, scale, is_causal, softmax_mode)``
+                returning ``(out, m, linv)`` all as float32.
+
+        Returns:
+            Concatenated output tensor in float32.
+        """
+        q_len = q.shape[-2]
+        kv_len = k.shape[-2]
+        prefix_len = kv_len - q_len
```
kamil-kaczor, line 282: Can we secure this so we run later with negative here?

vllm_gaudi/extension/ops.py
```diff
@@ -412,7 +412,9 @@ def _fsdpa_prompt_attention(query: torch.Tensor,
     recompute_mode = True
     assert attn_bias is not None or valid_seq_lengths is not None, \
         'Either attn_bias or valid_seq_lengths must be != None'
-    if is_causal and attn_bias is not None:
+
+    # moved to ModuleFusedSDPA and ModuleFP8FusedSDPA after the dispatching for sliced FusedSDPA when INC is not used
+    if is_causal and attn_bias is not None and os.getenv("QUANT_CONFIG") is not None:
```
kamil-kaczor: can we avoid calling os.getenv("QUANT_CONFIG") in every forward pass?

tests/unit_tests/test_fsdpa_slicing.py
```diff
@@ -0,0 +1,1404 @@
...
+                                          ctx_pad=pad,
+                                          graph_breaks=graph_breaks,
+                                          mode=mode)
+
+            cos_sim = torch.nn.functional.cosine_similarity(ref_out.flatten().float(),
+                                                            sliced_out.flatten().float(),
+                                                            dim=0).item()
+            max_abs_diff = (ref_out.float() - sliced_out.float()).abs().max().item()
+
+        assert cos_sim > 0.92, f"BF16 cosine similarity too low: {cos_sim}"
```
kamil-kaczor: isn't this a bit low? Maybe not 1 but arccos of 0.92 is like 23 degrees difference between vectors. The industry standard is like 0.999 but let's assume 3σ margin and make it 0.997. Would it pass also?

kamil-kaczor: Let me elaborate on that - because attention weights tokens based on their importance (how relevant, how close to what we want to generate) then we can even have a scenario where 50% of the tokens are garbage and we still get 0.92 cos similarity. xD It can happen when the least important part for the calculation is complete garbage (for example 50% starting tokens) and the rest is OK (the later 50% tokens). The less relevant part is discounted (example 10% of the calc weight) and more important part is weighted highly (90%) and that's why it can be so tricky to catch it. 0.997 is not perfect as it won't catch all isuess but will make sure that almost all tokens are OK

tests/unit_tests/test_fsdpa_slicing.py
```diff
@@ -0,0 +1,1404 @@
...
+                                          ctx_pad=pad,
+                                          graph_breaks=graph_breaks,
+                                          mode=mode)
+
+            cos_sim = torch.nn.functional.cosine_similarity(ref_out.flatten().float(),
+                                                            sliced_out.flatten().float(),
+                                                            dim=0).item()
+            max_abs_diff = (ref_out.float() - sliced_out.float()).abs().max().item()
+
+        assert cos_sim > 0.74, f"FP8 cosine similarity too low: {cos_sim}"
```
kamil-kaczor: I suggest 0.95 when comparing vs bf16 and 0.99 when comparing fp8 vs fp8

vllm_gaudi/extension/utils.py
```diff
@@ -198,13 +523,12 @@ def __init__(self, fusedSDPA):
         self.d_scale_q = torch.tensor(1.0, dtype=torch.float32)
         self.d_scale_k = torch.tensor(1.0, dtype=torch.float32)
         self.d_scale_v = torch.tensor(1.0, dtype=torch.float32)
+        self.d_scale_output = torch.tensor(1.0, dtype=torch.float32)
+        self._sliced_module = SlicedFP8FusedSDPA(parent=self)
 
     def quant_input(self, x, scale):
         return torch.ops.hpu.cast_to_fp8_v2(x, scale, False, False, torch.float8_e4m3fn)[0]
 
-    def dequant_output(self, output, scale):
```
kamil-kaczor, line 205: I know it's dead to us but I hope NIC doesn't need it. If there are fails after merging you can come back to this to check it as one of the culprits.

vllm_gaudi/extension/utils.py
```diff
@@ -221,9 +545,27 @@ def forward(
         window_size=None,
     ):
 
-        qinput = self.quant_input(query, self.scale_q)
```
kamil-kaczor, line 224: It can affect non-slicing path but if tests are OK we should be good. Another place to note for debug if something happens

vllm_gaudi/extension/utils.py
```diff
@@ -149,12 +151,246 @@ def forward(self, input, other, **kwargs):
...
+            # the causal part
+            for kv_chunk_idx in range(0, num_q_chunks - q_chunk_idx):
+                kv_start = prefix_len + q_end - (kv_chunk_idx + 1) * self.chunk_size
+                kv_start = max(kv_start, prefix_len)
+                kv_end = prefix_len + q_end - kv_chunk_idx * self.chunk_size
+                kv_chunk_size = kv_end - kv_start
+                k_chunk = k[..., kv_start:kv_end, :].contiguous()
+                v_chunk = v[..., kv_start:kv_end, :].contiguous()
+
+                is_causal_chunk = kv_chunk_idx == 0 and q_chunk_idx >= self.num_padded_query_chunks
```
kamil-kaczor: Not my idea but AI reviewer says:
"Finding 1. Diagonal chunks can lose both mask AND causality — vllm_gaudi/extension/utils.py, _chunked_attention

When num_padded_query_chunks == 0 (possible with low VLLM_PROMPT_QUERY_BUCKET_PAD_MAX) AND chunk size isn't 1024-aligned, then is_causal_chunk=False AND mask_chunk=None. The kernel computes full attention on a diagonal chunk — attending to future tokens. Low probability with defaults but one env var override away from silently producing wrong results."
and
"Missing test: chunk size doesn't evenly divide sequence length — would exercise the "first chunk is smaller" edge case that interacts with Finding 1"

This deep into the algo it's difficult for me to perfectly evaluate this finding without taking much more time understanding the edge cases. Please evaluate if it's valuable and if not just leave a comment.

kamil-kaczor (changes requested):

kamil-kaczor:
also please resolve conflicts

docs/configuration/env_variables.md
yangulei: It's for causal attention only, the padding must be 'right' for my understanding. Actually, the following code fixed padding to 'right':
https://github.com/vllm-project/vllm-gaudi/blob/7ea0676790dfa1652750bb52119ff2b7d7438ebe/vllm_gaudi/extension/ops.py#L377-L392
And the slicing for-loop assuming 'right' padding as it starts from the bottom-right of the attention mask, and pass attention mask for the first few chunks.

tests/unit_tests/test_fsdpa_slicing.py
yangulei: Done, thx.

tests/unit_tests/test_fsdpa_slicing.py
yangulei: Fixed, thx.

tests/unit_tests/test_fsdpa_slicing.py
yangulei: Original idea is to isolate the UTs. Fixed now, thx.

tests/unit_tests/test_fsdpa_slicing.py
yangulei: Fixed, thx.

vllm_gaudi/extension/utils.py
yangulei: Done, thx.

vllm_gaudi/extension/ops.py
yangulei: Done, thx.

vllm_gaudi/extension/utils.py
yangulei: Noted, thx.

vllm_gaudi/extension/utils.py
yangulei: Noted, thx.

vllm_gaudi/extension/utils.py
yangulei: num_padded_query_chunks should not be 0 as it is calculated by `ceil(VLLM_PROMPT_QUERY_BUCKET_PAD_MAX/VLLM_HPU_FSDPA_SLICE_CHUNK_SIZE)`. Added an assertion for `VLLM_PROMPT_QUERY_BUCKET_PAD_MAX >= block_size` to make it safer.

tests/unit_tests/test_fsdpa_slicing.py
yangulei: Yes you are right. The accuracy checking for FusedSDPA is tricky. The uniform random Q/K/V is used in the UT, cause nearly uniform near-zero attention. Which makes the fp8 quantization noise cannot be ignored. While the end to end lm_eval longbench accuracy is ok. I'm working on more realistic UT.

tests/unit_tests/test_fsdpa_slicing.py
yangulei: Done. The `cons_sim` is 0.999 for bf16 and 0.99 for fp8 after https://github.com/vllm-project/vllm-gaudi/pull/1155/changes/8e7679b6ac0e1e71f2d7010f7081b05c377deb72 and https://github.com/vllm-project/vllm-gaudi/pull/1155/changes/3fb1310f0383df33052d320aef11711a51f74866

docs/configuration/env_variables.md
kamil-kaczor: Weird but OK. We support padding both ways but I don't think it makes any difference

kamil-kaczor (approved):
lgtm

PR #1169, opened 2026-03-17
Add real context length to the high-level profile

yangulei:


yangulei:
The new high-level profile will show the `utils: decode_ctx_len` and `utils: prompt_ctx_len` as shown bellow:
image


And with the [helper script](https://github.com/vllm-project/vllm-gaudi/blob/aice/scripts/high_level_profile_to_csv.py) which converts the high-level profile to a `.csv` file, the `ctx_len` and the `bucket_ctx_len` columns show the padding in the `context` dimension as shown bellow:
image

PR #1177, opened 2026-03-18
Optimize the recipe launch time for FP8 Qwen3

yiliu30:
Split the input scale/weight for each expert to avoid explicit slice in MOE op which can cause performance regression.


adobrzyn:
@yiliu30 is it ready for review? Can you provide more info what this PR fixes?

adobrzyn (approved):

yiliu30:
Hi @adobrzyn could u help land it, thanks!

PR #1182, opened 2026-03-18
Optimize multimodal warmup: 17x faster via compile-only mode and .tolist() caching

moshehoori:
## Summary

Reduces multimodal warmup from **362s to 17s** on Qwen3-VL-32B-FP8 (1×Gaudi2, max_model_len=32768).
Overall warmup drops from **634s to 297s (53% reduction)**.

### 1. Cache `.tolist()` result across attention layers (`hpu_mm_encoder_attention.py`)

`HpuMMEncoderAttention._forward_sdpa` calls `(cu_seqlens[1:] - cu_seqlens[:-1]).tolist()` once per transformer layer. The vision encoder has 27+ layers, all using the **same** `cu_seqlens` tensor in a single forward pass. Each `.tolist()` forces a synchronous D2H copy (~0.35s blocking while waiting for all pending HPU ops).

**Fix:** Module-level cache keyed by `id(cu_seqlens)` — first layer computes and caches, remaining 26 layers reuse the result.

Also adds padding-aware split logic to handle cases where the query tensor is padded to a bucket size (`query.shape[2] > sum(lens)`), which occurs during compile-only warmup with FakeTensors.

**Impact:** CPU time per multimodal bucket drops from 10.31s to 0.22s (benefits both warmup and inference).

### 2. Compile-only mode for multimodal warmup (`hpu_model_runner.py`)

Graph warmup already uses `PT_COMPILE_ONLY_MODE` to skip HPU execution — only Synapse recipe compilation is needed. This PR extends the same approach to multimodal warmup by wrapping `embed_multimodal()` in `bc.env_setting("PT_COMPILE_ONLY_MODE", True)`.

Falls back gracefully to normal execution if compile-only fails on the first bucket.

### 3. CPU/HPU pipeline overlap (`hpu_model_runner.py`)

Splits `_prepare_dummy_scenario` into `_build_dummy_scenario` (CPU-only) and `_execute_dummy_scenario` (HPU). Uses a background thread to prepare the next bucket while HPU processes the current one. Applied to `warmup_graphs()`, `warmup_unified_graphs()`, and `warmup_multimodal_graphs()`.

Also parallelizes prompt/decode bucket generation.

## Results (Qwen3-VL-32B-Instruct-FP8, 1×Gaudi2)

| Phase | Before | After | Speedup |
|-------|--------|-------|---------|
| Multimodal (35 buckets) | 362s | 17s | **21x** |
| Sampler + Defrag | 9s | 9s | 1x |
| Graph warmup (168 buckets) | 245s | 234s | 1.05x |
| **Total warmup** | **634s** | **297s** | **2.1x** |

## Verification

Tested on Qwen3-VL-32B-Instruct-FP8 with max_model_len=32768, max_num_seqs=16:
- ✅ Warmup completes without errors (all 35 MM buckets in compile-only mode)
- ✅ Text inference: correct output
- ✅ Multimodal inference: correctly describes test image after warmup

104 min warmup for Qwen3-VL-235B

### DIFF
--- a/vllm_gaudi/ops/hpu_mm_encoder_attention.py
+++ b/vllm_gaudi/ops/hpu_mm_encoder_attention.py
@@ -2,6 +2,13 @@ import torch
 import torch.nn.functional as F
 from vllm.model_executor.layers.attention.mm_encoder_attention import MMEncoderAttention
 
+# Cache for cu_seqlens → lens conversion to avoid repeated
+# .tolist() D2H syncs across attention layers (each sync ~0.35s).
+# A single vision encoder forward pass calls _forward_sdpa 27+ times
+# with the SAME cu_seqlens tensor — caching saves ~9s per bucket.
+_lens_cache_id = None
+_lens_cache_val = None
+
 
 @MMEncoderAttention.register_oot()
 class HpuMMEncoderAttention(MMEncoderAttention):
@@ -17,6 +24,7 @@ class HpuMMEncoderAttention(MMEncoderAttention):
         (batch_size x seq_len x hidden_size) or
         (batch_size x seq_len x num_heads x head_size)
         """
+        global _lens_cache_id, _lens_cache_val
         bsz, q_len = query.size()[:2]
         kv_len = key.size(1)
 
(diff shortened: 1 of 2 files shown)


iboiko-habana:
@moshehoori please rebase this PR if it is still needed. Otherwise PR will be closed. Thanks

PR #1192, opened 2026-03-19
Share kv cache tensor with attention for granite

michalkuligowski:


jbyczkow (commented):
Please verify as_strided performance impact

PR #1198, opened 2026-03-19
Prefix caching support for HPUMambaMixer2

jbyczkow:
Implement prefix caching for the Mamba2 hybrid model (Granite 4.0-h) on Gaudi HPUs. Changes:

- Add prefix caching metadata computation in HPUModelRunner for both prefill and decode paths: block index tracking, chunk-to-block mappings, and SSM state load/store indices
- Modify HPUMambaMixer2 to use cached Mamba SSM/conv states from the block table when prefix caching is enabled
- Refactor causal_conv1d_pytorch to support initial_states for cached convolution state restoration
- Extend HPU attention backends to carry Mamba prefix caching tensors through the attention metadata
- Pad PC metadata tensors to fixed sizes (max_cached_blocks) to avoid dynamic shape recompilation on Habana
- Pass enable_prefix_caching flag through to hpu_causal_conv1d_fn
- Fix offset computation in decode path for block index calculation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ Homepage = "https://github.com/vllm-project/vllm-gaudi"
 
 [tool.setuptools.packages.find]
 where = ["."]
-include = ["vllm_gaudi"]
+include = ["vllm_gaudi*"]
 
 [tool.yapf]
 column_limit = 120


adobrzyn:
**Finding 1** 🔴 **Critical** · `vllm_gaudi/v1/worker/hpu_model_runner.py` — outer else branch in `_form_prefill_batch`

`mamba_chunks_to_block_mapping` is not set to `None` in the outer `else` branch (when `self.num_mamba_layers == 0`), but it IS passed to `make_prefill_metadata` below. This will cause a `NameError` for non-mamba models.

The `else` block sets `blocks_caching_range = None` and `seqlens_offsets_for_blocks = None` but omits `mamba_chunks_to_block_mapping`.

**Suggestion:** Add `mamba_chunks_to_block_mapping = None` to the else branch alongside the other assignments:

```python
        else:
            prep_initial_states = None
            load_indices_tensor = None
            store_indices_tensor = None
            has_initial_states_p = None
            last_chunk_indices_p = None
            padding_mask_flat = None
            query_start_loc_p = None
            blocks_caching_range = None
            mamba_chunks_to_block_mapping = None  # <-- missing
            seqlens_offsets_for_blocks = None
```

---

PR #1204, opened 2026-03-20
[Granite-4h-small] Enable defragmentator

agrabow:
Exclude mamba layers from defragmentation. Identify ids of self-attention layers and include them in defragmentation.


michalkuligowski (changes requested):
Does this PR  expect that underlaying raw tensor will be shared between mamba nad attention or assumes separate memory for mamba and attention?

PR #1242, opened 2026-03-25
avoid using pip show to get habana-torch-plugin version

dtrifiro:
When `pip` is not installed, the `subprocess` call will fail, and the plugin will fail to initialize.

`importlib.metadata` is a cleaner way to get plugin version info.

### DIFF
--- a/vllm_gaudi/extension/environment.py
+++ b/vllm_gaudi/extension/environment.py
@@ -4,7 +4,10 @@
 # This source code is licensed under the Apache 2.0 license found in the
 # LICENSE file in the root directory of this source tree.
 ###############################################################################
+import importlib
 import os
+
+from contextlib import suppress
 
 from .logger import logger
 from .config import Value, boolean, split_values_and_flags, Any, Disabled, Enabled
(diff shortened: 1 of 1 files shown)


vllm_gaudi/extension/environment.py
```diff
@@ -54,17 +54,13 @@ def _get_vllm_hash(_):
...
-                            text=True,
-                            stdout=subprocess.PIPE,
-                            stderr=subprocess.PIPE)
-    version_re = re.compile(r'Version:\s*(?P<version>.*)')
-    match = version_re.search(output.stdout)
-    if output.returncode == 0 and match:
-        return match.group('version')
+    import importlib
+
+    metadata = importlib.metadata.metadata("habana-torch-plugin")
```
adobrzyn: Won't it raise PackageNotFoundError and crashes whole thing?

dtrifiro: Yep. that's correct. Fixed it in the latest push

dtrifiro: @adobrzyn fyi

adobrzyn (approved):

PR #1290, opened 2026-04-02
Use finite numbers for the attention mask

yangulei:
Revert "Use Boolean attention mask (#1032)" and use finite numbers for attention mask to solve the accuracy issue for fp8 FusedSDPA.


yangulei:
Close as included in https://github.com/vllm-project/vllm-gaudi/pull/1285

PR #1320, opened 2026-04-08
Separate conv1d for Granite 4.0 (v0.17.1-style)

jbyczkow:
Granite 4.0 (hpu_mamba_mixer2) needs a simpler single-sequence conv1d implementation matching v0.17.1 behavior. The current causal_conv1d_pytorch.py was refactored in v0.19.0 to support batched prefill with prefix-caching metadata, which is not needed by Granite 4.0.

Changes:
- Add granite_causal_conv1d.py with granite_causal_conv1d_fn and granite_causal_conv1d_update — simplified conv1d that processes one sequence at a time (padded_batch == 1) without prefix-caching support.
- Update hpu_mamba_mixer2.py to import and call the Granite-specific conv1d functions instead of the generic hpu_causal_conv1d_fn/update.
- Remove unused variables (mamba_block_size, block_idx_* placeholders) from the Granite mixer forward path.

The existing causal_conv1d_pytorch.py is left intact for other models.

### DIFF
--- a/vllm_gaudi/ops/granite_causal_conv1d.py
+++ b/vllm_gaudi/ops/granite_causal_conv1d.py
@@ -1,0 +1,195 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+# Copyright (c) 2024, Tri Dao.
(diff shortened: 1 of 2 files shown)


wpyszka (approved):
approved for 0.19.0

PR #1323, opened 2026-04-08
Fix for proper KV cache slot addressing for Hybrid models

ksmusz:
## Summary

Fix KV cache slot addressing mismatch in prefill path for hybrid (Mamba + Attention) models such as Granite4.0.

## Problem

For hybrid models, the logical `block_size` is inflated beyond the HPU paged-attention kernel's native block size (128) during page-size unification with Mamba layers. After the Qwen3.5 enablement commit, `self.attn_block_size` was introduced to track the actual kernel block size and was correctly applied in the **decode** path. However, the **prefill** path continued using the larger `self.block_size` for slot index arithmetic.

This caused prefill and decode to compute different physical KV cache positions for the same token:

- **Prefill** wrote KV entries at `block_table[block_id] * block_size + offset` (using the inflated size)
- **Decode** read KV entries at `block_table[block_id] * attn_block_size + offset` (using the kernel size)

Since these addresses differ, decode reads stale or unrelated data from the KV cache, producing incorrect attention outputs.

Pure Transformer models are **not affected** because their `block_size` and `attn_block_size` are always equal (both 128 on HPU).

## Fix

Use `self.attn_block_size` (the HPU paged-attention kernel block size) instead of `self.block_size` (the logical/hybrid-inflated block size) for all prefill-path KV cache slot addressing. The decode path already uses `self.attn_block_size` via `decode_block_size` and is not changed.

Changed locations in `vllm_gaudi/v1/worker/hpu_model_runner.py` (5 hunks):

| Location | Change |
|----------|--------|
| `_get_new_batch_contents` | Block count for block_table indexing now uses `attn_block_size` |
| `_make_attn_bias` | `repeat_interleave` expansion of context groups now uses `attn_block_size` |
| `_form_prefill_batch` | `divmod` block assignment, `token_slots` computation, and `num_context_blocks` for block_table slicing now use `attn_block_size` |
| `_create_dummy_prefill_batch_contents` | Dummy prefill block count now uses `attn_block_size` |
| `_PAD_SLOT_ID` | Padding slot sentinel now uses `attn_block_size` |

Bucketing is kept on `self.block_size` so that file-based bucket configurations (generated at the original block size) continue to match. When the two sizes differ, `target_blocks` returned by the bucketing function is scaled from `block_size` units to `attn_block_size` units for correct context_blocks padding.


jbyczkow (approved):

wpyszka (approved):
approved for 0.19.0

PR #1324, opened 2026-04-08
Fix Synapse GC compile failure for FP8-quantized models

slokesha:
When serving FP8-quantized models with PT_HPU_LAZY_MODE=0, graph compilation fails with:
`verifyConvertScale: expected H2D convert scale tensor`

Fix

Skip any attribute listed in `mod.scale_members` — INC's per-module registry of scale tensor names, populated by register_scale() — inside `_move_remaining_tensors_to_device()`. All other stray tensors (MLA weights W_UK_T/W_UV, KV-quant range scalars, etc.) are still moved to HPU as before.

Affected models

Any model quantized via Intel Neural Compressor with scale_format: const: LLaMA, DeepSeek-R1, Qwen, Mixtral, and others.

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -249,6 +249,8 @@ def _move_remaining_tensors_to_device(model: torch.nn.Module, device: str) -> No
 
     moved = 0
     for mod in model.modules():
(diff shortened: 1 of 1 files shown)


iboiko-habana (approved):

slokesha:
@iboiko-habana Can we merge this?

iboiko-habana (approved):

PR #1326, opened 2026-04-09
fix: exclude non-attention caches from defragmenter for hybrid models (cherry-pick to 0.19.0)

kamil-kaczor:
Cherry-pick of #1294 to `releases/v0.19.0`.

## Problem

Enabling `VLLM_CONTIGUOUS_PA=true` + `VLLM_DEFRAG=true` causes **~32 GiB memory spikes** on hybrid models like Qwen3.5-9B (24 GDN/linear_attention + 8 full_attention layers), eventually crashing with OOM.

**Root cause:** `OnlineDefragmenter` receives ALL kv_caches including GDN/mamba state tensors. The defrag code uses per-token indexing (`block_id * block_size`), which is only valid for `FullAttentionSpec` KV caches. For GDN state tensors, these indices are wildly out of bounds, causing massive temporary memory allocations on HPU.

## Fix

Filter `kv_caches` passed to `OnlineDefragmenter` to include **only** `FullAttentionSpec` layer caches. Non-hybrid models are unaffected.

### Cherry-picked commits:
- `1580798` fix: exclude non-attention caches from defragmenter for hybrid models
- `b0122c3` fix: resolve mypy errors for defragmenter type annotations
- `5afcd18` Fix pre-commit

Fixes: GAUDISW-247457

### DIFF
--- a/vllm_gaudi/extension/defragmentation.py
+++ b/vllm_gaudi/extension/defragmentation.py
@@ -27,7 +27,7 @@ class CacheSwapUtils(torch.nn.Module):
         self.enable_prefix_caching = get_config().prefix_caching
         self.block_slots = torch.arange(0, block_size, dtype=torch.long, device=device)
 
-    def forward(self, srcs: torch.tensor, dsts: torch.tensor, caches: list[torch.tensor], block_size: int):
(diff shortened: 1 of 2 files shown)


kamil-kaczor:
Closing as it's too risky to merge it to 0.19. Can reopen if decided otherwise

PR #1328, opened 2026-04-09
enable GPT-OSS-120B single-card inference

osavchenkox:
Store MoE expert weights as packed MXFP4 in HBM instead of dequantizing to BF16 at load time. A single shared BF16 workspace (~6 GB) is allocated and reused across all 36 MoE layers for runtime dequantization, reducing weight memory from and enabling inference on a single card.

### DIFF
--- a/vllm_gaudi/extension/ops.py
+++ b/vllm_gaudi/extension/ops.py
@@ -651,8 +651,13 @@ class VllmMixtureOfExpertsOp(VllmMixtureOfExpertsOpBase):
         """Build and cache weight/bias *views only* (no torch.stack / no extra allocation)."""
         experts_range = range(self.num_experts)
 
-        self._cached_w13_views = tuple(self.w13_list[i].weight.squeeze() for i in experts_range)
-        self._cached_w2_views = tuple(self.w2_list[i].weight.squeeze() for i in experts_range)
+        # MXFP4 path: weights are not set in MoeMatmul (dequantized at runtime)
+        if hasattr(self.w13_list[0], 'weight') and self.w13_list[0].weight is not None:
+            self._cached_w13_views = tuple(self.w13_list[i].weight.squeeze() for i in experts_range)
+            self._cached_w2_views = tuple(self.w2_list[i].weight.squeeze() for i in experts_range)
+        else:
+            self._cached_w13_views = None
+            self._cached_w2_views = None
 
         # optional bias views (no copy)
         if self.bias is not None:
--- a/vllm_gaudi/ops/hpu_fused_moe.py
+++ b/vllm_gaudi/ops/hpu_fused_moe.py
@@ -31,12 +31,26 @@ from vllm.model_executor.layers.fused_moe.router.routing_simulator_router import
 from vllm_gaudi.extension.ops import (VllmMixtureOfExpertsOp)
 from vllm_gaudi.extension.runtime import get_config
 from vllm.model_executor.utils import set_weight_attrs
+from vllm.model_executor.layers.quantization.utils.ocp_mx_utils import OCP_MX_BLOCK_SIZE
 from vllm_gaudi.utils import has_quant_config
 from vllm_gaudi.v1.worker.hpu_dp_utils import dispatch_hidden_states, dispatch_tensor, get_hpu_dp_metadata
 
 
 def _normalize_moe_activation(activation):
     return activation.value if isinstance(activation, Enum) else activation
+
+
+def dequant_mxfp4_hpu(packed, scales, *, out):
+    """Dequantize MXFP4 packed weights on HPU into a pre-allocated output tensor.
+
+    Args:
+        packed: uint8 tensor of shape (E, rows, groups, block_half) — packed FP4 blocks
+        scales: uint8 tensor of shape (E, rows, groups) — per-group scales
+        out: pre-allocated BF16 tensor of shape (E, rows, groups * block_half * 2)
+    """
+    from vllm_gaudi.models.gptoss_mxfp4 import convert_moe_packed_tensors
+    result = convert_moe_packed_tensors(packed, scales, dtype=out.dtype)
+    out.copy_(result)
 
 
 @UnquantizedFusedMoEMethod.register_oot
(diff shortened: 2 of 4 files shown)


SKRohit:
@osavchenkox  maybe we can modify `convert_moe_packed_tensors` and pass `layer._shared_w2_workspace` to it, so that result could be stored directly into `layer._shared_w2_workspace`.

adobrzyn:
@osavchenkox is this ready for review? Please solve conficts

vllm_gaudi/ops/hpu_fused_moe.py
```diff
@@ -111,32 +145,50 @@ def create_weights(self, layer: torch.nn.Module, num_experts: int, hidden_size:
...
-            w13_bias = torch.nn.Parameter(torch.zeros(num_experts,
-                                                      2 * round_up(intermediate_size_per_partition, 32),
-                                                      dtype=params_dtype),
-                                          requires_grad=False)
+            assert hidden_size % OCP_MX_BLOCK_SIZE == 0, \
+                f"hidden_size ({hidden_size}) must be divisible by OCP_MX_BLOCK_SIZE ({OCP_MX_BLOCK_SIZE})"
+            assert intermediate_size_per_partition % OCP_MX_BLOCK_SIZE == 0, \
+                f"intermediate_size_per_partition ({intermediate_size_per_partition}) must be divisible by " \
+                f"OCP_MX_BLOCK_SIZE ({OCP_MX_BLOCK_SIZE})"
+            groups_h = hidden_size // OCP_MX_BLOCK_SIZE
```
michalkuligowski, line 153: If intermediate_size_per_partition is not divisible by 32, the assert intermediate_size_per_partition % OCP_MX_BLOCK_SIZE == 0 (OCP_MX_BLOCK_SIZE=32) already guarantees it, so in practice they match. Still, replace the bias sizing with plain 2 * intermediate_size_per_partition (and hidden_size for w2_bias) for consistency, or keep round_up for both packed and bias. Today the biases are padded and the packed weight is not — this could hide bugs if the divisibility assumption is ever weakened.

vllm_gaudi/extension/ops.py
```diff
@@ -651,8 +651,13 @@ def _cache_weight_lists(self):
         """Build and cache weight/bias *views only* (no torch.stack / no extra allocation)."""
         experts_range = range(self.num_experts)
 
-        self._cached_w13_views = tuple(self.w13_list[i].weight.squeeze() for i in experts_range)
-        self._cached_w2_views = tuple(self.w2_list[i].weight.squeeze() for i in experts_range)
+        # MXFP4 path: weights are not set in MoeMatmul (dequantized at runtime)
+        if hasattr(self.w13_list[0], 'weight') and self.w13_list[0].weight is not None:
```
michalkuligowski, line 655: MoeMatmul is a plain torch.nn.Module and does not define a default weight attribute, so hasattr is False until set_weight() runs — correct. But calling hasattr(module, 'weight') can trigger lazy attribute hooks. Safer and clearer:

w = getattr(self.w13_list[0], 'weight', None)
if w is not None:

vllm_gaudi/ops/hpu_fused_moe.py
```diff
@@ -184,6 +236,32 @@ def apply_monolithic(
...
+                              out=layer._shared_w2_workspace)
+
+            E = layer.local_num_experts
+            w13_list = tuple(layer._shared_w13_workspace[i] for i in range(E))
+            w2_list = tuple(layer._shared_w2_workspace[i] for i in range(E))
+
+            output = torch.ops.hpu.mixture_of_experts.bias_fused_weights(
+                hidden_states=x,
+                expert_routing_table=topk_ids,
+                router_weights=topk_weights,
```
michalkuligowski, line 254: Double BF16 cast in forward_oot MXFP4 branch - the non-MXFP4 gpt_oss branch a few lines down passes topk_weights.to(x.dtype) / topk_ids.to(torch.int64). Conversely, apply_monolithic's new MXFP4 branch omits the .to(int64) cast even though topk_ids was cast above unconditionally to int64 when not layer.use_grouped_topk. For EP-routing grouped-topk paths (use_grouped_topk=True), topk_ids may still be int32/other and will enter the kernel uncast. Unify: add .to(torch.int64) / .to(x.dtype) in both MXFP4 branches to match the existing defensive casts, or verify grouped-topk is not a valid config for gpt_oss+MXFP4.

vllm_gaudi/extension/ops.py
```diff
@@ -651,8 +651,13 @@ def _cache_weight_lists(self):
         """Build and cache weight/bias *views only* (no torch.stack / no extra allocation)."""
```
michalkuligowski, line 651: The MXFP4 runtime-dequant path in apply_monolithic/forward_oot, and the new convert_moe_packed_tensors(out=…) behavior, are uncovered. At minimum add a CPU-runnable unit test for convert_moe_packed_tensors comparing the out=-preallocated path vs. the legacy allocating path on small synthetic inputs (this catches shape/dtype regressions on any dev machine, not just HPU). Ideally also a smoke test under tests/full_tests/ using a small MXFP4 gpt-oss-style card.

vllm_gaudi/ops/hpu_fused_moe.py
```diff
@@ -39,6 +40,20 @@ def _normalize_moe_activation(activation):
     return activation.value if isinstance(activation, Enum) else activation
 
 
+@torch.compiler.disable
+def dequant_mxfp4_hpu(packed, scales, *, out):
```
michalkuligowski, line 44: dequant_mxfp4_hpu is @torch.compiler.disable'd, but the workspace is shared across all 36 MoE layers. For HPU Graph capture (if enabled), each layer's MoE forward writes the entire workspace before reading it, so it must be outside a captured graph or the graph must include the write. Need to verify this doesn't break use_hpu_graphs=True / PT_HPU_LAZY_MODE paths. A torch.hpu.synchronize() or htorch.core.mark_step() may be required between dequant and kernel call to ensure ordering on lazy mode — worth an explicit test.

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -574,11 +574,46 @@ def _init_mamba_split_weights(model):
             module._init_split_weights()
 
 
+def _init_mxfp4_shared_workspace(model):
```
michalkuligowski, line 577: As flagged in copilot review: torch.empty(..., device=ref.w13_packed.device) will allocate ~6 GB wherever w13_packed lives. If model.to("hpu") is skipped (fake-HPU tests, dry-run init), this becomes a 6 GB CPU allocation. Guard with if device.type != 'hpu': return or check is_fake_hpu().

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -574,11 +574,46 @@ def _init_mamba_split_weights(model):
...
+    from vllm.model_executor.layers.fused_moe import FusedMoE
+    from vllm.model_executor.layers.quantization.utils.ocp_mx_utils import OCP_MX_BLOCK_SIZE
+
+    moe_layers = [m for m in model.modules()
+                  if isinstance(m, FusedMoE) and hasattr(m, 'w13_packed')]
+    if not moe_layers:
+        return
+
+    # All MoE layers have identical shapes — derive dimensions from the first
+    ref = moe_layers[0]
```
michalkuligowski, line 592: _init_mxfp4_shared_workspace derives (E, I2, H, ...) only from moe_layers[0]. If a future model mixes MoE layers with different expert counts or intermediate sizes (rare but legal), subsequent layers will stride into the wrong shape. Add a simple assert loop:

> for l in moe_layers[1:]: 

>         assert l.w13_packed.shape == ref.w13_packed.shape
>         assert l.w2_packed.shape  == ref.w2_packed.shape

PR #1333, opened 2026-04-09
Adding image rebuild step for missing images in CI

ksmusz:
## Problem

When a single test job fails and is re-run in isolation via the GitHub Actions UI ("Re-run failed jobs" or "Re-run this job"), the Docker image built by `pre_merge_hpu_test_build` is no longer present on the runner. The `cleanup_docker_image` job runs with `if: always()` at the end of every workflow run and removes it. As a result, re-running any downstream test job would immediately fail with a Docker image-not-found error, even though the test itself was never the root cause of the failure.

## Solution

Added a **"Rebuild Docker image if not present"** step (preceded by a repository checkout) to all 7 test jobs:

- `hpu_unit_tests`
- `hpu_pd_tests`
- `hpu_perf_tests`
- `hpu_dp_tests`
- `e2e`
- `calibration_tests`
- `calibration_arg_parsing_tests`

Each step checks whether the expected Docker image exists on the runner before proceeding:

```bash
if ! docker image inspect "$IMAGE_NAME" > /dev/null 2>&1; then
    # rebuild using the vLLM commit cached in pre_merge_hpu_test_build outputs
    docker build --no-cache --build-arg VLLM_COMMIT_ARG=<commit> ...
else
    echo "Image already present, skipping rebuild."
fi
```

The `VLLM_COMMIT_ARG` is taken from `needs.pre_merge_hpu_test_build.outputs.target_commit`, which GitHub Actions caches from the original run — so the rebuild always uses the exact same vLLM commit as the original build, ensuring test reproducibility.

## Image Lifecycle / Pruning Update

To avoid rebuilding/deleting the same image repeatedly across multiple reruns while still preventing long-term disk growth, we added two controls:

1. **Image labels on all CI build paths** (both initial build and fallback rebuilds):
   - `ci.owner=vllm-gaudi-pre-merge`
   - `ci.head_sha=<PR head sha>`

2. **TTL-based janitor prune** in `pre_merge_hpu_test_build`:

```bash
docker image prune -f \
  --filter "label=ci.owner=vllm-gaudi-pre-merge" \
  --filter "until=24h"
```

This keeps recently built images available for reruns (fast path), and removes stale labeled images once they are older than 24 hours.

## Behavior

| Scenario | Behavior |
|---|---|
| Normal full run | Image exists → rebuild step is a no-op (~1 second) |
| Single job re-run after cleanup | Image missing → rebuild triggered automatically |
| Multiple job re-runs (sequential runner) | First re-run rebuilds; subsequent ones skip (image already present) |
| Old cached images | Pruned when a later CI run executes and image age is `>24h` |

## Notes

- The runner is single-concurrency (self-hosted), so no race conditions between parallel rebuilds.
- The repository checkout step added before the rebuild is required so `Dockerfile.ci` is available on disk.
- TTL pruning is eligibility-based: images are deleted by the next CI execution that runs the prune step (not by a background daemon at exactly 24h).
- No changes to CI flow, job dependencies, or test logic.


ksmusz:
@copilot resolve the merge conflicts in this pull request

iboiko-habana:
Closing at this moment

PR #1334, opened 2026-04-09
Fix Synapse GC compile failure for FP8-quantized models

jiminha:
Pull changes from main (https://github.com/vllm-project/vllm-gaudi/pull/1324)

GC compile error(verifyConvertScale: expected H2D convert scale tensor) was introduced after https://github.com/vllm-project/vllm-gaudi/pull/1183 for all the models quantized via Intel Neural Compressor with scale_format: const: LLaMA, DeepSeek-R1, Qwen, Mixtral, and other.  This is due to we are also moving all scale tensors. This change is to skip them.


iboiko-habana (approved):

wpyszka (approved):
fix, approved for 0.19.0

PR #1335, opened 2026-04-09
Fix Qwen3.5 compile error for upstream change

jiminha:
Fix Qwen3.5 compile error  + Enable Qwen3.5 e2e lm_eval test.


jiminha:
THis will be included in https://github.com/vllm-project/vllm-gaudi/pull/1312

PR #1341, opened 2026-04-13
Finetuning bucket rules for 128k hybrid model edge cases

ksmusz:


mgawarkiewicz-intel:
it won't get into 0.19.0

PR #1346, opened 2026-04-14
skip HPU graphs for long (query + context) prefills

yangulei:
## Motivation
Current implementation use the following logic to skip the HPU graphs for long prefills:

https://github.com/vllm-project/vllm-gaudi/blob/bcff6c8a4e41dae81bbfd762961430f7607637f9/vllm_gaudi/v1/worker/hpu_model_runner.py#L3105-L3106

While:
- `self.max_cudagraph_capture_size` is not set by default,
- the batched tokens is calculated by `batch_size * seq_len` which miss the context length which is comparable or even larger than the query length as the chunked-prefill and APC is enabled by default.

Those lead to unnecessary HPU graphs for the compute-bound long prefills and introduce much more memory footprint which may cause OOM crash.

## Changes:
- Set the `self.max_cudagraph_capture_size` to `self.max_num_batched_tokens` if it is not set,
- Includes the context tokens while calculating the batched tokens.

> This is an re-implementation of the reverted PR https://github.com/vllm-project/vllm-gaudi/pull/780.

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -1207,14 +1207,15 @@ class HPUModelRunner(HpuKVConnectorModelRunnerMixin):
         self.use_hpu_graph = not self.model_config.enforce_eager
         self.max_batch_size = self.scheduler_config.max_num_seqs
         self.max_num_seqs = self.scheduler_config.max_num_seqs
-        self.max_cudagraph_capture_size = self.vllm_config.compilation_config.max_cudagraph_capture_size
         if prompt_profile_cfg:
             self.max_prefill_batch_size = prompt_profile_cfg[0]
         else:
             self.max_prefill_batch_size = with_default(get_config().VLLM_PROMPT_BS_BUCKET_MAX, 1)
         self.seen_configs: set = set()
-        self.max_num_batched_tokens = \
-            self.scheduler_config.max_num_batched_tokens
+        self.max_num_batched_tokens = self.scheduler_config.max_num_batched_tokens
+        self.max_cudagraph_capture_size = self.vllm_config.compilation_config.max_cudagraph_capture_size
+        if self.max_cudagraph_capture_size is None:
+            self.max_cudagraph_capture_size = self.max_num_batched_tokens
         self.use_prefix_caching = (self.vllm_config.cache_config.enable_prefix_caching)
         self.bucketing_manager = HPUBucketingManager()
         max_num_prefill_seqs = self.max_num_seqs if self.use_merged_prefill \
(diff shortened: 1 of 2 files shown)


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -4386,8 +4385,17 @@ def _compile_region(self, model, name, module):
...
-        return not self.model_config.enforce_eager
+    def _use_graphs(self, attn_metadata=None, batch_size=0):
+        if self.model_config.enforce_eager:
+            return False
+        # skip HPU graphs for long (query + context) prefills
+        if attn_metadata is not None and attn_metadata.is_prompt:
+            seq_len = attn_metadata.seq_len()
+            num_blocks = attn_metadata.num_blocks()
+            total_tokens = (batch_size * seq_len + num_blocks * attn_metadata.block_size)
+            if total_tokens > self.max_cudagraph_capture_size:
```
michalkuligowski, line 4593: For debuggability, it might be helpful to add a debug/info log when graphs are skipped for a long prefill (e.g., logger.debug("Skipping HPU graphs for prefill: total_tokens=%d > max_capture=%d", total_tokens, self.max_cudagraph_capture_size)). This would help users understand why certain prefills aren't using graphs, especially when diagnosing performance. The existing code at line 3119 does include graphs{'T' if use_graphs else 'F'} in the profiler event name, which partially addresses this.

yangulei: done, thanks

michalkuligowski (approved):

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -4539,8 +4538,20 @@ def _compile_region(self, model, name, module):
     def _compile(self, module):
         return torch.compile(module, **self.compile_config.get_compile_args())
 
-    def _use_graphs(self):
-        return not self.model_config.enforce_eager
+    def _use_graphs(self, attn_metadata=None, batch_size=0):
```
kamil-kaczor: why batch_size=0 as default? Someone can forget this and get weird values

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -4539,8 +4538,20 @@ def _compile_region(self, model, name, module):
...
+        if self.model_config.enforce_eager:
+            return False
+        # skip HPU graphs for long (query + context) prefills
+        if attn_metadata is not None and attn_metadata.is_prompt:
+            seq_len = attn_metadata.seq_len()
+            num_blocks = attn_metadata.num_blocks()
+            total_tokens = (batch_size * seq_len + num_blocks * attn_metadata.block_size)
```
kamil-kaczor, line 4592: I don't understand this. Isn't total tokens num_blocks * block_size (with padding included)? Same for batch_size * seq_len

kamil-kaczor (changes requested):

vllm_gaudi/v1/worker/hpu_model_runner.py
yangulei: Fixed by removing the default values, thanks.

vllm_gaudi/v1/worker/hpu_model_runner.py
yangulei: No, `num_blocks * block_size` is the total **context** tokens, and `batch_size * seq_len` is the total **query** tokens. Take the chunked prefill for a prompt with `10240` tokens, `max_num_batch_tokens=8192` and `block_size=128` as an example:
- The prefill `[bs, seq_len, num_blocks]` for the first chunk will be `[1, 8192, 0]`.
- For the second chunk it will be `[1, 2048, 8192/128=64]`, where `2048` is the query length (`q_len` in FSDPA) and`8192` is the context length (`kv_len - q_len` in FSDPA).

kamil-kaczor: I see, ty

kamil-kaczor (approved):
lgtm

PR #1351, opened 2026-04-14
Accept PEP 440 versions in build detection

### LINKED ISSUE #1350: Build version detection rejects PEP 440 post-release versions from habana-torch-plugin
## Description

The build version detection in `vllm_gaudi/extension/environment.py` uses `pip show habana-torch-plugin` to get the Gaudi SW build version, then validates it against the regex pattern `^\d+\.\d+\.\d+\.\d+$`.

However, `habana-torch-plugin` can report PEP 440 post-release versions like `1.23.0.post1`, which don't match this pattern. This causes vllm-gaudi to fail at startup with:

```
RuntimeError: build: '1.23.0.post1' doesn't match pattern '^\d+\.\d+\.\d+\.\d+$'!
You can override detected build by specifying VLLM_BUILD env variable
```

## Root Cause

In `environment.py`, the `_get_build()` function returns the raw version string from `pip show`, but the `regex` validator only accepts `X.Y.Z.W` format (four dot-separated integers). PEP 440 post-release suffixes like `.postN` are valid Python package versions but are rejected by this pattern.

https://github.com/vllm-project/vllm-gaudi/blob/main/vllm_gaudi/extension/environment.py#L108

## Impact

Any environment where `habana-torch-plugin` reports a `.postN` version will fail to start vllm-gaudi without manually setting the `VLLM_BUILD` env var as a workaround.

## Proposed Fix

Normalize the version string in `_get_build()` to convert PEP 440 post-release versions (e.g., `1.23.0.post1`) into the expected four-segment format (e.g., `1.23.0.1`) before returning. This way the existing `VersionRange()` comparisons in `features.py` continue to work correctly.

wjhrdy:
## Summary

`habana-torch-plugin` can report PEP 440 versions like `1.23.0.post1`. The build version regex `^\d+\.\d+\.\d+\.\d+$` rejects anything that isn't four dot-separated integers, causing startup failure:

```
RuntimeError: build: '1.23.0.post1' doesn't match pattern '^\d+\.\d+\.\d+\.\d+$'!
```

## Fix

Replace the hardcoded regex with a `pep440_version()` validator that uses `packaging.version.Version` (already a dependency via `config.py`) to validate. This accepts any valid PEP 440 version — post-releases, dev releases, release candidates — without enumerating suffixes in a regex.

## Changes

- `validation.py`: Add `pep440_version()` checker
- `environment.py`: Use `pep440_version()` instead of `regex()` for build version validation

`VersionRange()` in `features.py` already uses `packaging.version.Version` to parse the build string, so this makes validation consistent with consumption.

Fixes #1350

Signed-off-by: Willy Hardy <whardy@redhat.com>

### DIFF
--- a/vllm_gaudi/extension/environment.py
+++ b/vllm_gaudi/extension/environment.py
@@ -11,7 +11,7 @@ from contextlib import suppress
 
 from .logger import logger
 from .config import Value, boolean, split_values_and_flags, Any, Disabled, Enabled
-from .validation import choice, regex
+from .validation import choice, pep440_version
 
 _VLLM_VALUES = {}
 
@@ -105,8 +105,7 @@ def get_environment():
         Value('build',
               _get_build,
               env_var_type=str,
-              check=regex(r'^\d+\.\d+\.\d+\.\d+$',
-                          hint='You can override detected build by specifying VLLM_BUILD env variable')),
+              check=pep440_version(hint='You can override detected build by specifying VLLM_BUILD env variable')),
         Value('engine_version', _get_vllm_engine_version, env_var_type=str),
         Value('bridge_mode', _get_pt_bridge_mode, env_var_type=str, check=choice('eager', 'lazy')),
         VllmValue('model_type', str),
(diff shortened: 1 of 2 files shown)


adobrzyn:
Hi @wjhrdy, please fix precommit issues so the CI can run :)

adobrzyn (approved):

PR #1352, opened 2026-04-15
[GAUDISW-247384] Add pre-merge-trigger.yaml

bmyrcha:
---

⚠️ **Branch-protection coordination required**

With this change, worker jobs (`pre-commit`, `pre_merge_hpu_test_build`, `pre_merge_hpu_test`, `check_dockerfile_changes`, …) now run via `workflow_dispatch` and **no longer appear as required status checks on PRs**. The only check surfacing on PRs will be `Pre-merge trigger / execute_pre_merge`.

Branch-protection rulesets on `main` must be updated before/at merge: replace the old required contexts with `Pre-merge trigger / execute_pre_merge`.

### DIFF
--- a/.github/workflows/pre-merge-trigger.yaml
+++ b/.github/workflows/pre-merge-trigger.yaml
@@ -1,7 +1,6 @@
 name: Pre-merge trigger
 
 on:
-  workflow_dispatch: {}
   pull_request_target:
     types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
     branches:
--- a/.github/workflows/pre-merge-trigger.yaml
+++ b/.github/workflows/pre-merge-trigger.yaml
@@ -19,6 +19,7 @@ concurrency:
 jobs:
   execute_pre_merge:
     runs-on: ubuntu-latest
+    timeout-minutes: 720
     permissions:
       actions: write       # dispatch workflows, read run status, cancel orphaned runs
       pull-requests: read  # read PR labels for skip-gaudi-tests check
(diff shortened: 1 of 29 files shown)


.github/workflows/pre-merge.yaml
```diff
@@ -1,26 +1,25 @@
...
+      is_merge_group: { description: 'Triggered from merge queue (true/false)', required: true, type: string, default: 'false' }
 
 permissions:
   pull-requests: write
   # This line allows it to read the status of checks like DCO
   checks: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ inputs.head_sha || github.ref }}
```
kamil-kaczor:
```suggestion
  group: ${{ github.workflow }}-${{ inputs.pr_number || inputs.head_sha || github.ref }}
```

.github/workflows/pre-merge-trigger.yaml
```diff
@@ -0,0 +1,121 @@
+name: Pre-merge trigger
+
+on:
+  workflow_dispatch: {}
```
kamil-kaczor: is it for debug?

bmyrcha: It is not functional now as currently this workflow needs github.event payload and probably workflow_dispatch can be safely remove. If we would like to keep it, we'd need to add inputs and adjust workflow to use it.

I'll remove it for now

kamil-kaczor (approved):
lgtm

PR #1360, opened 2026-04-16
Eliminate Llama4 torch.compile recompilations on HPU

adobrzyn:
Llama4 Maverick has 48 heterogeneous decoder layers (NoPE/RoPE, MoE/MLP, ChunkedLocal/regular attention) causing torch._dynamo guard recompilations on the first benchmark run, resulting in P99 TTFT ~9081ms vs ~939ms on subsequent runs.

Changes:
- Add HpuLlama4ForConditionalGeneration model registry override
- Initialize residual as zeros instead of None to eliminate type guard
- Apply branch-free attention (boolean masks + torch.where) to remove Python if/else guards on nope/rotary_emb/temperature_tuning
- Unify ChunkedLocalAttention -> Attention to eliminate type guards
- Add validation warmup for heterogeneous layer guard specializations

Note: _mark_unbacked_dim0 for block_list already added by PR #1329.

Reduces first-run P99 TTFT from ~9081ms to ~1037ms (88.6% improvement).

### DIFF
--- a/vllm_gaudi/models/__init__.py
+++ b/vllm_gaudi/models/__init__.py
@@ -44,6 +44,10 @@ def register_model():
     from vllm_gaudi.models.qwen3_moe import HpuQwen3MoeForCausalLM  # noqa: F401
     ModelRegistry.register_model("Qwen3MoeForCausalLM", "vllm_gaudi.models.qwen3_moe:HpuQwen3MoeForCausalLM")
 
+    from vllm_gaudi.models.llama4 import HpuLlama4ForConditionalGeneration  # noqa: F401
+    ModelRegistry.register_model("Llama4ForConditionalGeneration",
+                                 "vllm_gaudi.models.llama4:HpuLlama4ForConditionalGeneration")
+
     import vllm_gaudi.models.deepseek_v2  # noqa: F401
 
     from vllm_gaudi.models.deepseek_ocr import HpuDeepseekOCRForCausalLM  # noqa: F401
--- a/vllm_gaudi/models/llama4.py
+++ b/vllm_gaudi/models/llama4.py
@@ -16,7 +16,6 @@ This module provides:
      to eliminate torch._dynamo type guards across layers.
 """
 
-import logging
 import types
 from itertools import islice
 
@@ -25,6 +24,7 @@ import torch
 import torch.nn as nn
 from vllm.config import VllmConfig
 from vllm.distributed import get_pp_group
+from vllm.logger import init_logger
 from vllm.model_executor.models.llama4 import (
     Llama4ForCausalLM as UpstreamLlama4ForCausalLM,
     Llama4Model as UpstreamLlama4Model,
@@ -33,7 +33,7 @@ from vllm.model_executor.models.mllama4 import (
     Llama4ForConditionalGeneration as UpstreamLlama4ForConditionalGeneration, )
 from vllm.sequence import IntermediateTensors
 
-logger = logging.getLogger(__name__)
+logger = init_logger(__name__)
 
 
 class HpuLlama4Model(UpstreamLlama4Model):
@@ -91,17 +91,17 @@ class HpuLlama4ForCausalLM(UpstreamLlama4ForCausalLM):
         self.model.__class__ = HpuLlama4Model
 
         # Apply branch-free attention patches (only in compile mode)
+        # Note: _unify_feed_forward_types is deferred to post-load via
+        # apply_hpu_llama4_post_load_patches() to avoid breaking weight loading
+        # (wrapping changes named_parameters() keys and hides .experts attribute).
         if not htorch.utils.internal.is_lazy():
             layers = getattr(self.model, "layers", [])
             _apply_branch_free_attention(layers)
             unified = _unify_attention_types(layers)
-            ff_unified = _unify_feed_forward_types(layers)
             logger.info(
                 "HpuLlama4ForCausalLM: applied branch-free attention patches, "
-                "unified %d ChunkedLocalAttention -> Attention, "
-                "unified %d feed_forward modules",
+                "unified %d ChunkedLocalAttention -> Attention",
                 unified,
-                ff_unified,
             )
 
 
@@ -122,6 +122,11 @@ def _apply_branch_free_attention(layers):
         logger.warning("No RoPE layer found to reference rotary_emb from")
         return
 
+    # If no layer has qk_norm (use_qk_norm=False), install nn.Identity()
+    # so the branchfree forward can call self.qk_norm() unconditionally.
+    if ref_qk_norm is None:
+        ref_qk_norm = nn.Identity()
+
     patched = 0
     for layer in layers:
         attn = layer.self_attn
@@ -137,12 +142,15 @@ def _apply_branch_free_attention(layers):
 def _patch_attention_module(attn, ref_rotary_emb, ref_qk_norm):
     """Patch a single Llama4Attention module to be branch-free."""
     if attn.rotary_emb is None:
-        attn.rotary_emb = ref_rotary_emb
+        # Use object.__setattr__ to avoid nn.Module registering a shared
+        # reference as a submodule under multiple parent paths, which would
+        # confuse tools that walk named_modules() (quantization, FX tracing).
+        object.__setattr__(attn, "rotary_emb", ref_rotary_emb)
 
     # QK norm: install reference on NoPE layers that have None
     has_qk_norm = attn.qk_norm is not None
     if not has_qk_norm and ref_qk_norm is not None:
-        attn.qk_norm = ref_qk_norm
+        object.__setattr__(attn, "qk_norm", ref_qk_norm)
 
     device = next(attn.parameters()).device
     attn.register_buffer(
@@ -153,7 +161,7 @@ def _patch_attention_module(attn, ref_rotary_emb, ref_qk_norm):
     attn.register_buffer(
         "_apply_temp_tuning",
         torch.tensor(
-            getattr(attn, "attn_temperature_tuning", False) and attn.nope,
+            getattr(attn, "attn_temperature_tuning", False),
             dtype=torch.bool,
             device=device,
         ),
@@ -208,6 +216,11 @@ def _unify_attention_types(layers):
 
     Since ChunkedLocalAttention does NOT override forward(), the __class__
     swap is behaviorally identical. get_kv_cache_spec is preserved as instance method.
+
+    WARNING: After this swap, isinstance(x, ChunkedLocalAttention) returns False.
+    Currently safe because maybe_set_chunked_attention_layers uses string matching
+    on backend names. If upstream ever switches to isinstance checks, this will
+    need updating. A _was_chunked_local marker is set for future detection.
     """
     from vllm.model_executor.layers.attention.attention import Attention
     from vllm.model_executor.layers.attention.chunked_local_attention import (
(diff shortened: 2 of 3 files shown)


kamil-kaczor:
**Finding 1** 🔴 **Critical** · `vllm_gaudi/models/llama4.py:241-253`

`_unify_feed_forward_types` runs in `HpuLlama4ForConditionalGeneration.__init__`, which executes before `default_loader.load_weights()`. After wrapping, MoE params become `…feed_forward.inner.experts.w13_weight` in `named_parameters()`, but upstream `Llama4Model.load_moe_expert_weights` builds `full_param_name` from the checkpoint key (no `.inner.`) and does `params_dict[full_param_name]` → KeyError. Also accesses `self.layers[i].feed_forward.experts.expert_map` directly → AttributeError on the wrapper.

**Suggestion:** Do the wrapping after `load_weights` (e.g. via `apply_model_specific_patches` hook), or override `__getattr__` on `_HpuLlama4FeedForward` to forward to `self.inner` and strip `.inner.` from state_dict.

---

kamil-kaczor:
**Finding 2** 🔴 **Critical** · `vllm_gaudi/models/llama4.py:163-203`

`_branchfree_attention_forward` calls `self.qk_norm(q_for_norm.float())` unconditionally, but `_apply_branch_free_attention` only early-returns when `ref_rotary_emb is None` — not `ref_qk_norm`. When `config.use_qk_norm == False`, upstream sets `self.qk_norm = None` and this path hits `TypeError: 'NoneType' object is not callable`.

**Suggestion:** Install `nn.Identity()` when `ref_qk_norm is None`, or guard the norm block at patch time.

---

kamil-kaczor:
**Finding 3** 🟡 **Suggestion** · `vllm_gaudi/models/llama4.py:212-230`

`_unify_attention_types` does `attn_inner.__class__ = Attention`. Any `isinstance(x, ChunkedLocalAttention)` check in upstream will silently return False. Current `maybe_set_chunked_attention_layers` uses string matching on backend names so it works today, but a future upstream switch to `isinstance` would silently disable chunked-local attention.

**Suggestion:** Set a marker `attn_inner._was_chunked_local = True` and add a comment documenting the hazard.

---

kamil-kaczor:
**Finding 4** 🟡 **Suggestion** · `vllm_gaudi/models/llama4.py:60-72`

`HpuLlama4Model.forward` reimplements the aux-hidden-states loop inline. Upstream calls `self._maybe_add_hidden_state(aux_hidden_states, idx + 1, hidden_states, residual)` and records an entry at idx=0 before the loop. Eagle3 speculative decoding will produce a different aux tensor set on HPU vs upstream.

**Suggestion:** Call `self._maybe_add_hidden_state(...)` — only the `residual = None → zeros_like(...)` line needs to diverge.

---

kamil-kaczor:
**Finding 5** 🟡 **Suggestion** · `vllm_gaudi/models/llama4.py:139-147`

`_patch_attention_module` installs the same `ref_rotary_emb` / `ref_qk_norm` objects on many layers via `attn.rotary_emb = …`. `nn.Module.__setattr__` registers them as submodules each time → same Python object under N parent paths in `named_modules()` / `state_dict()`. Tools that walk submodules (quantization, FX tracing) will double-process.

**Suggestion:** Use `object.__setattr__(attn, "rotary_emb", ref_rotary_emb)` to bypass nn.Module registration.

---

kamil-kaczor:
**Finding 6** 🟡 **Suggestion** · `vllm_gaudi/v1/worker/hpu_model_runner.py:5296-5311`

Validation warmup runs unconditionally when `_has_heterogeneous_layers` is set, adding 1–2 prompt buckets + 1 decode bucket of compile time with no env gate. After it runs, `self.graphed_buckets = saved_graphed` discards tracking. Relies on `warmup_graphs` being idempotent w.r.t. compile cache — otherwise the real warmup recompiles these buckets.

**Suggestion:** Add `VLLM_DISABLE_VALIDATION_WARMUP` env switch; verify validation-warmed buckets aren't recompiled on a real run, else keep them in `graphed_buckets`.

---

kamil-kaczor:
**Finding 7** 🟡 **Suggestion** · `vllm_gaudi/models/llama4.py:285-294`

`is_hpu_llama4_model` branches on `HpuLlama4ForCausalLM`, but that class is defined and never registered with `ModelRegistry` (only `HpuLlama4ForConditionalGeneration` is). Text-only `Llama4ForCausalLM` falls back to upstream and this branch never fires.

**Suggestion:** Register `HpuLlama4ForCausalLM` in `vllm_gaudi/models/__init__.py`, or drop the dead branch and unused class.

---

kamil-kaczor:
**Finding 8** 🟢 **Nit** · `vllm_gaudi/models/llama4.py:135-138`

`_apply_temp_tuning = (attn_temperature_tuning AND nope)` duplicates a condition — upstream `Llama4Attention.__init__` already sets `self.attn_temperature_tuning = self.nope and config.attn_temperature_tuning`.

**Suggestion:** `torch.tensor(getattr(attn, "attn_temperature_tuning", False), …)`; drop the redundant `and attn.nope`.

---

kamil-kaczor:
**Finding 9** 🟢 **Nit** · `vllm_gaudi/models/llama4.py`

Uses `logging.getLogger(__name__)` while the rest of the repo uses `init_logger` from `vllm_gaudi.logger`. Inconsistent log format/level.

**Suggestion:** `from vllm_gaudi.logger import init_logger; logger = init_logger(__name__)`.

---

kamil-kaczor:
**Finding 10** 🟢 **Nit** · PR-level

No unit test covers the monkey-patching (layer-0 vs layer-1 output parity, class-swap KV cache spec, `_HpuLlama4FeedForward` state_dict round-trip). Given how invasive these patches are, a small test under `tests/unit_tests/` would catch regressions from upstream changes.

**Suggestion:** Add a small unit test under `tests/unit_tests/` exercising the patched layers.

---

adobrzyn:
**Re: Finding 4** (aux-hidden-states loop) — No action needed.

`_maybe_add_hidden_state` does **not exist** in upstream vLLM. The upstream `LlamaModel.forward()` (which `Llama4Model` inherits without overriding) uses the exact same inline pattern:

```python
for idx, layer in enumerate(islice(self.layers, self.start_layer, self.end_layer)):
    if idx in self.aux_hidden_state_layers:
        aux_hidden_states.append(hidden_states + residual)
```

Our `HpuLlama4Model.forward()` is a faithful copy with only the `residual = torch.zeros_like()` divergence. There is no upstream helper to call.

---
**Re: Finding 6** (validation warmup env gate) — No action needed.

The validation warmup **is the core mechanism** that eliminates the 9081→1037ms first-run TTFT regression. Adding a disable switch would let users silently re-introduce the problem. The `graphed_buckets = saved_graphed` reset is intentional — we only want to trigger guard specializations during validation, not mark these buckets as already warmed (the real warmup handles them with correct/full bucket sets). The torch.compile cache is idempotent, so the real warmup does not recompile these buckets.

---
**Re: Finding 10** (unit test) — Agreed, but deferred as a follow-up.

The monkey-patching is integration-tested via the end-to-end CI (which passed). A dedicated unit test exercising patched layers would be valuable but is better as a separate follow-up task to keep this PR focused on the performance fix.

---
*Findings 1, 2, 3, 5, 7, 8, 9 have been addressed in commit 889c27a.*

kamil-kaczor:
### Re-review after latest changes

Thanks for the updates! Status of previous findings:

| # | Prev | Status | Note |
|---|------|--------|------|
| 1 | 🔴 | ✅ Fixed | Wrapping moved to `apply_hpu_llama4_post_load_patches()` called after `load_weights()` |
| 2 | 🔴 | ✅ Fixed | `nn.Identity()` fallback when `qk_norm is None`; mask still gates output |
| 3 | 🟡 | 🟡 Partial | `_was_chunked_local = True` marker + WARNING docstring added |
| 4 | 🟡 | ❌ Open | Still reimplements aux-hidden-states loop; see new Finding N1 below |
| 5 | 🟡 | ✅ Fixed | Uses `object.__setattr__` to bypass `nn.Module` registration |
| 6 | 🟡 | 🟡 Partial | Gated on `_has_heterogeneous_layers`; no env kill-switch |
| 7 | 🟡 | ❌ Open | `HpuLlama4ForCausalLM` still defined but unregistered |
| 8 | 🟢 | ✅ Fixed | |
| 9 | 🟢 | ✅ Fixed | Uses `init_logger` |
| 10 | 🟢 | ❌ Open | No tests added |

4 new findings posted below.

---

kamil-kaczor:
### New findings from re-review

**Finding N1** 🔴 **Critical** · `vllm_gaudi/models/llama4.py:50-78`

`HpuLlama4Model.forward` captures `hidden_states + residual` at loop-entry for `idx ∈ [0, N)`. Upstream `LlamaModel.forward` captures at `idx=0` pre-loop and then at `idx+1` post-layer, so the post-last-layer state is captured when `N ∈ aux_hidden_state_layers`. This PR never emits the final layer's output — any Eagle3 config specifying the final layer index silently loses that tensor.

**Suggestion:** call `self._maybe_add_hidden_state(...)` (preferred), or replicate the pre-loop `idx=0` + `idx+1` post-layer pattern.

---

**Finding N2** 🟡 **Suggestion** · `vllm_gaudi/v1/worker/hpu_model_runner.py:953`

`self._has_heterogeneous_layers: bool = False` is set True only inside `apply_model_specific_patches`. If a runner instance is reused across model loads (some test harnesses do this), a previous Llama4 run leaves the flag True. Harmless today, brittle.

**Suggestion:** reset to `False` at the top of `load_model`.

---

**Finding N3** 🟢 **Nit** · `vllm_gaudi/models/llama4.py:227-229`

`from vllm.model_executor.layers.attention.attention import Attention` is imported inside a function called unconditionally from `__init__`. If upstream renames the module path, `__init__` crashes before weights load with a confusing traceback.

**Suggestion:** top-level import, optionally guarded with try/except and a clearer error message.

---

**Finding N4** 🟢 **Nit** · `vllm_gaudi/models/llama4.py:87-104` vs `264-283`

`HpuLlama4ForCausalLM.__init__` and `HpuLlama4ForConditionalGeneration.__init__` bodies are near-identical. Extract an `_apply_hpu_patches(self, model_root)` helper to avoid drift, regardless of how Finding 7 is resolved.

---

adobrzyn:
### Re-review response

**Finding N1** (Critical — aux-hidden-states) — **No action needed.**

The reviewer states upstream captures at idx=0 pre-loop and at idx+1 post-layer. This is incorrect. Upstream `LlamaModel.forward()` ([llama.py L420-428](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/llama.py#L420-L428)) uses the **exact same** inline pattern as our code:

```python
aux_hidden_states = []
for idx, layer in enumerate(islice(self.layers, self.start_layer, self.end_layer)):
    if idx in self.aux_hidden_state_layers:
        aux_hidden_states.append(hidden_states + residual)
    hidden_states, residual = layer(positions, hidden_states, residual, **extra_layer_kwargs)
```

There is no pre-loop capture, no post-layer idx+1 pattern, and `_maybe_add_hidden_state` does not exist in upstream. `Llama4Model` does not override `forward()`. Our code is a verbatim copy with only `residual = torch.zeros_like()` changed.

---

**Finding N2** (Suggestion — reset flag on reload) — **Not needed.**

`_has_heterogeneous_layers` is set unconditionally in `apply_model_specific_patches()` which runs every time a model is loaded (`model_runner._has_heterogeneous_layers = is_hpu_llama4_model(model_runner.model)`). A stale `True` from a previous load is always overwritten. No risk of incorrect state.

---

**Finding N3** (Nit — function-level import) — **Not needed.**

The `Attention` / `ChunkedLocalAttention` import paths are core vLLM infrastructure that hasn't changed in 2+ years. A function-level import is idiomatic for lazy loading and avoids circular import risk. If upstream renames these paths, the crash at init time is the correct failure mode — it surfaces immediately during model loading.

---

**Finding N4** — ✅ **Fixed** in commit 7e84a4a.

Extracted `_apply_hpu_llama4_init_patches(model_root)` shared helper used by both `HpuLlama4ForCausalLM.__init__` and `HpuLlama4ForConditionalGeneration.__init__`.

---

kamil-kaczor (approved):
lgtm

PR #1361, opened 2026-04-16
fix: stabilize decode block_list shape for long-context workloads

adobrzyn:
## Problem

With non-contiguous paged attention, the flattened `block_list` (sum of all per-request blocks) can exceed the bucketing system's `_fallback_max_ctx` cap. For example, 16 requests at 200k context = ~25,000 block entries, but the cap is only `max_blocks * 3 ≈ 13,464`.

Since `pad_list` only pads (never truncates), the tensor keeps its original length — a unique shape every decode step, causing HPU graph recompilation (~5-10s) on every single token.

## Fix

In `get_habana_paged_attn_buffers` (non-contiguous PA path), when `block_bucket_size < len(block_list)`, round up `len(block_list)` via `calc_fallback_value` with `base_step=32`. This produces coarse bucket sizes with ~900 steps of headroom, reducing recompilations from every step to roughly every ~900 steps.

## Changes
- `vllm_gaudi/v1/worker/hpu_model_runner.py`: Add `calc_fallback_value` import and guard in the non-contiguous PA path
- `tests/unit_tests/test_bucketing.py`: Add `test_calc_fallback_value_stabilizes_oversized_block_list` verifying shape stability across consecutive decode steps

## Validation
- All 39 bucketing unit tests pass
- `calc_fallback_value(25008, 32)` = 25,920 — stable for 200 consecutive decode steps (vs 0 without fix)
- ruff check clean

### DIFF
--- a/tests/unit_tests/test_bucketing.py
+++ b/tests/unit_tests/test_bucketing.py
@@ -681,3 +681,32 @@ def test_padding_aware_decode_cfgs_contiguous_pa_clamps_block_range(mock_get_con
                                                max_blocks=3593)
 
     assert block_cfg == [3465, 128, 3593, 899, 25]
+
+
+def test_calc_fallback_value_stabilizes_oversized_block_list():
+    """GAUDISW-247865: decode block_list that exceeds the max bucket must be
+    rounded up via calc_fallback_value so that the padded shape is stable
+    across adjacent lengths (no recompilation storm)."""
+
+    base_step = 32
+
+    # 1. Bug scenario: 16 reqs × 1563 blocks at 200k ctx with block_size=128
+    oversized = 25008
+    bucket = calc_fallback_value(oversized, base_step)
+    assert bucket >= oversized, (
+        f"Fallback bucket {bucket} must cover the actual block_list length {oversized}")
+    assert bucket % base_step == 0, (
+        f"Fallback bucket {bucket} must be divisible by base_step {base_step}")
+
+    # 2. Shape stability: adjacent lengths must map to the same bucket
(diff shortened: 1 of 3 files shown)


tests/unit_tests/test_bucketing.py
```diff
@@ -190,24 +192,28 @@ def __init__(self, **kwargs):
...
+    max_model_len = 91964
+    _, _, block_cfg = strategy.get_decode_cfgs(max_num_seqs=max_num_seqs,
                                                block_size=block_size,
                                                max_num_batched_tokens=131072,
-                                               max_model_len=91964,
+                                               max_model_len=max_model_len,
                                                max_blocks=max_blocks)
 
-    expected_max = max_blocks * 3  # 10779
+    import math
```
kamil-kaczor, line 214: we have import math as _math above and here import math

tests/unit_tests/test_bucketing.py
```diff
@@ -681,3 +690,29 @@ def test_padding_aware_decode_cfgs_contiguous_pa_clamps_block_range(mock_get_con
                                                max_blocks=3593)
 
     assert block_cfg == [3465, 128, 3593, 899, 25]
+
+
+def test_calc_fallback_value_stabilizes_oversized_block_list():
+    """GAUDISW-247865: decode block_list that exceeds the max bucket must be
```
kamil-kaczor, line 696: We shouldn't use GAUDISW-number in our codebase/PRs etc

kamil-kaczor (changes requested):

PR #1362, opened 2026-04-16
Bucketing edge cases finetune for longer ctx

ksmusz:
### Problem

When using `VLLM_BUCKETING_FROM_FILE` with `max_model_len=131072` and a non-power-of-two `block_size` (e.g. 528 for Granite hybrid models), the `not_over_max_model_len` filter silently rejected the max context bucket (ctx=249). The old check `(query + ctx * block_size) <= max_model_len` failed because the padded bucket query (e.g. 8192) plus full context (249×528=131,472) exceeded 131,072 — even though at runtime the actual last-chunk query is shorter than the bucket size.

This caused two cascading failures:
1. Missing max-ctx bucket forced fallback bucket generation at runtime.
2. The fallback (`calc_fallback_value`) overshoots dramatically (e.g. ctx=235 → 448 blocks), triggering OOM during graph compilation.

Additionally, `mamba_decode_corrector` used `math.floor` to cap decode ctx, producing 1984 instead of the correct 1992 (`8 × ceil(131072/528)`). This caused repeated fallback warnings during decode.

### Changes

**`vllm_gaudi/extension/bucketing/common.py`**:
- Relaxed `not_over_max_model_len` to check `ctx <= ceil((max_model_len - 1) / block_size)` — only context blocks matter, not the padded query size.
- Added prompt-specific fallback cap in `generate_fallback_bucket` to limit ctx to `ceil(max_model_len / block_size)`, preventing oversized fallback buckets.
- Fixed `mamba_decode_corrector` (`math.floor` → `math.ceil`) and removed it as it became identical to `correct_for_max_model_len`. Simplified `get_corrector` accordingly.

**`vllm_gaudi/v1/worker/hpu_model_runner.py`**:
- Fixed `_prepare_dummy_scenario` to cap warmup dummy total tokens when `query + ctx * block_size > max_model_len` in the non-merged-prefill path, recomputing context blocks so that `scheduled_tokens == prompt_query_len` exactly.

### DIFF
--- a/vllm_gaudi/extension/bucketing/common.py
+++ b/vllm_gaudi/extension/bucketing/common.py
@@ -235,7 +235,7 @@ class HPUBucketingManager():
 
     ### RETRIEVE BUCKETS FUNCTIONS ###
 
-    def generate_fallback_bucket(self, batch_size, seq_len, ctx):
+    def generate_fallback_bucket(self, batch_size, seq_len, ctx, is_prompt=False):
         assert self.max_num_batched_tokens is not None
         new_batch_size = calc_fallback_value(batch_size, self.fallback_bs_base_step)
         if new_batch_size > self.max_num_seqs:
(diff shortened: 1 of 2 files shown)


vllm_gaudi/extension/bucketing/common.py
```diff
@@ -249,19 +249,30 @@ def generate_fallback_bucket(self, batch_size, seq_len, ctx):
...
-            # Safety cap: limit to max prepared decode bucket ctx to prevent
-            # catastrophic graph compilation from corrupted batch data.
-            if self._fallback_max_ctx > 0 and new_ctx > self._fallback_max_ctx:
-                logger().warning(f"Fallback ctx {new_ctx} exceeds max prepared "
-                                 f"decode bucket ctx {self._fallback_max_ctx}, capping.")
-                new_ctx = self._fallback_max_ctx
+            if is_prompt:
+                # For prompt buckets, cap to the theoretical max number of
+                # context blocks derived from max_model_len.
+                max_prompt_ctx = math.ceil(self.max_model_len / self.block_size)
```
michalkuligowski: please fix copilot issue:

max_prompt_ctx = math.ceil((self.max_model_len - 1) / self.block_size)

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -4977,12 +4977,21 @@ def _prepare_dummy_scenario(self, prompt_cfg, decode_cfg):
...
+                        prompt_total_tokens, prompt_num_context_blocks = \
+                             self.get_merged_prefill_seq_lens(prompt_query_len,
+                                                         prompt_num_blocks)
+                    else:
+                        # With chunked prefill, query bucket + ctx * block_size
+                        # can exceed max_model_len when the bucket pads beyond
+                        # the actual last chunk size. Recompute context blocks
+                        # so that ctx * block_size + query_len <= max_model_len
+                        # and scheduled tokens == prompt_query_len exactly.
+                        capped_ctx = ((self.max_model_len - prompt_query_len) // self.block_size)
```
michalkuligowski: please fix copilot review:

capped_ctx = max(0, (self.max_model_len - prompt_query_len) // self.block_size)

vllm_gaudi/extension/bucketing/common.py
ksmusz: Sure, I'll fix this

vllm_gaudi/v1/worker/hpu_model_runner.py
ksmusz: I've been thinking on this case a bit and how real the problem is. I've came to conclusion, that I will implement the mentioned negative value safeguard, but I'm not sure if the case here would actually propagate up here it wouldn't fail either way (but before my change it would too). Here's the analysis for the future:

`prompt_query_len > max_model_len` can only happen when `max_num_batched_tokens > max_model_len`, which occurs in the non-chunked-prefill path (upstream enforces `max_num_batched_tokens >= max_model_len` when chunked prefill is disabled). The `else` branch we're in here only runs without merged prefill, and this is the exact config where query buckets can exceed `max_model_len`.

However, `max(0, ...)` doesn't actually fix it — it shifts the failure downstream:

**Example:** `max_model_len=512, max_num_batched_tokens=2048, block_size=128`, bucket `(1, 2048, 0)`:

```
capped_ctx = max(0, (512 - 2048) // 128) = 0
prompt_total_tokens = 0 * 128 + 2048 = 2048
```

We'd then create a dummy request with `total_tokens=2048` and `scheduled_tokens=2048` for a 512-token model — that may fail in KV cache / attention downstream.

The real fix would be to not generate unreachable buckets where `query > max_model_len` at bucket generation time, but that's a pre-existing issue outside this PR's scope (this PR targets chunked prefill with long contexts, where `max_num_batched_tokens < max_model_len` always holds and `query > max_model_len` cannot happen).

Previously we'd generate such oversized buckets too, so it may be a case for analysis of our bucketing logic.

jbyczkow (approved):

PR #1364, opened 2026-04-16
Logging omitted buckets when bucketing from file is enabled

ksmusz:
## Warn users about omitted buckets from custom bucketing file

When a user specifies custom buckets via `VLLM_BUCKETING_FROM_FILE`, some buckets may be silently dropped by internal filters (e.g. exceeding `max_model_len` or `max_num_batched_tokens`). Previously there was no indication that requested buckets were not warmed up, which could lead to unexpected fallback compilations at runtime.

This change adds a warning log listing each omitted bucket and the filter condition that rejected it, so users are informed when their custom bucket configuration is not fully honored. The warning only applies to file-based buckets — auto-generated buckets remain unchanged.


jbyczkow (approved):

kamil-kaczor (approved):
lgtm

PR #1368, opened 2026-04-16
Move mm graph warmup under pt compile only context

shepark:
# Changes
* Add _get_dummy_mm_inputs_with_options() helper: Uses the multimodal processor's get_dummy_processor_inputs() with custom mm_options (width/height/num_frames) and processor.apply() to generate dummy inputs that actually reflect the target resolution.
* Fix _get_mm_dummy_batch(): Replace the dead-code mm_options construction and the MultiModalRegistry().get_dummy_mm_inputs() call with the new helper, so each resolution bucket gets correctly-sized dummy data.
* Simplify warmup_multimodal_graphs() warmup_lists construction: Clean up the config-driven width/height extraction logic.
* Move MM warmup inside PT_COMPILE_ONLY_MODE context in warmup_model(): This avoids actual tensor computation during warmup, reducing warmup time.
* Remove unused MultiModalRegistry import.
* Call processor once and reuse across all resolution buckets.
* Run MM warmup under PT_COMPILE_ONLY_MODE only in torch.compile
  mode; skip in lazy mode to avoid GC errors.
* Use dummy mask in compile-only mode to avoid data-dependent crash

### DIFF
--- a/vllm_gaudi/models/qwen2_5_vl.py
+++ b/vllm_gaudi/models/qwen2_5_vl.py
@@ -35,6 +35,7 @@ from vllm.model_executor.models.utils import (maybe_prefix, cast_overflow_tensor
 from vllm.multimodal.inputs import MultiModalFieldConfig
 
 import habana_frameworks.torch.core as htcore
+import habana_frameworks.torch.internal.bridge_config as bridge_config
 from habana_frameworks.torch.hpex.kernels import FusedSDPA
 
 logger = init_logger(__name__)
@@ -411,8 +412,22 @@ class Qwen2_5_VisionTransformerStaticShape(Qwen2_5_VisionTransformer):
                     0.0)
             rot_pos_emb_sin = F.pad(rot_pos_emb_sin, (0, 0, 0, num_pad_tokens), "constant", 0.0)
 
-            padding_attn_mask_full = create_block_diagonal_attention_mask(cu_seqlens)
-            padding_attn_mask_window = create_block_diagonal_attention_mask(cu_window_seqlens)
+            # In PT_COMPILE_ONLY_MODE, Synapse compiles graph recipes without
+            # executing tensors, so cu_seqlens values are uninitialized garbage.
+            # create_block_diagonal_attention_mask() reads cu_seqlens values to
+            # determine mask shapes and control flow (e.g. .to("cpu") fallback
+            # when seq_len > 40000), which is unsafe during compile-only warmup.
+            # Since create_block_diagonal_attention_mask() has no guard for the
+            # warmup path despite pt_compile_only_mode being active, we bypass
+            # it here with a dummy mask of the correct [bucket_size, bucket_size]
+            # shape, which is sufficient for FusedSDPA recipe compilation.
+            if bridge_config.get_pt_compile_only_mode():
+                mask_size = int(bucket_size)
+                padding_attn_mask_full = torch.zeros(mask_size, mask_size, dtype=torch.bool, device=self.device)
+                padding_attn_mask_window = torch.zeros(mask_size, mask_size, dtype=torch.bool, device=self.device)
+            else:
+                padding_attn_mask_full = create_block_diagonal_attention_mask(cu_seqlens)
+                padding_attn_mask_window = create_block_diagonal_attention_mask(cu_window_seqlens)
 
             # static part
             htcore.mark_step()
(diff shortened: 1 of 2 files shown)


adobrzyn:
**Finding 1** 🟡 **Medium** · `hpu_model_runner.py:L5323-L5332`

The new `warmup_lists` loop picks width/height from the first modality config that exists regardless of whether that modality's warmup is enabled. Old code gated on `mm_limits[modality] != limit_value`. Now if image config exists but warmup is disabled, video dims are never checked. Also, `break` fires when `mm_options` is truthy even if width/height are None, preventing fallthrough to video.

**Suggestion:** Gate the loop on `is_image_warmup`/`is_video_warmup` and move `break` inside the `if width is not None and height is not None:` block.

---

adobrzyn:
**Finding 2** 🟡 **Medium** · `hpu_model_runner.py:L5237-L5251`

The new helper uses `self.model_config` instead of the previous `self.model_config_copy` (a `deepcopy` created at init for multimodal use). If the processor/registry caches state inside the config, this could cause side effects on the shared config — the deep copy existed to prevent exactly this.

**Suggestion:** Use `self.model_config_copy` in `_get_dummy_mm_inputs_with_options` and `_get_mm_warmup_processor`, or document why `self.model_config` is now safe.

---

adobrzyn:
**Finding 2** 🟡 **Medium** · `hpu_model_runner.py:L5237-L5251`

The new helper uses `self.model_config` instead of the previous `self.model_config_copy` (a `deepcopy` created at init for multimodal use). If the processor/registry caches state inside the config, this could cause side effects on the shared config — the deep copy existed to prevent exactly this.

**Suggestion:** Use `self.model_config_copy` in `_get_dummy_mm_inputs_with_options` and `_get_mm_warmup_processor`, or document why `self.model_config` is now safe.

---

adobrzyn:
**Finding 2** 🟡 **Medium** · `hpu_model_runner.py:L5237-L5251`

The new helper uses `self.model_config` instead of the previous `self.model_config_copy` (a `deepcopy` created at init for multimodal use). If the processor/registry caches state inside the config, this could cause side effects on the shared config — the deep copy existed to prevent exactly this.

**Suggestion:** Use `self.model_config_copy` in `_get_dummy_mm_inputs_with_options` and `_get_mm_warmup_processor`, or document why `self.model_config` is now safe.

---

adobrzyn:
**Finding 2** 🟡 **Medium** · `hpu_model_runner.py:L5237-L5251`

The new helper uses `self.model_config` instead of the previous `self.model_config_copy` (a `deepcopy` created at init for multimodal use). If the processor/registry caches state inside the config, this could cause side effects on the shared config — the deep copy existed to prevent exactly this.

**Suggestion:** Use `self.model_config_copy` in `_get_dummy_mm_inputs_with_options` and `_get_mm_warmup_processor`, or document why `self.model_config` is now safe.

---

adobrzyn:
**Finding 2** 🟡 **Medium** · `hpu_model_runner.py:L5237-L5251`

The new helper uses `self.model_config` instead of the previous `self.model_config_copy` (a `deepcopy` created at init for multimodal use). If the processor/registry caches state inside the config, this could cause side effects on the shared config — the deep copy existed to prevent exactly this.

**Suggestion:** Use `self.model_config_copy` in `_get_dummy_mm_inputs_with_options` and `_get_mm_warmup_processor`, or document why `self.model_config` is now safe.

---

adobrzyn:
**Finding 3** 🟡 **Medium** · `hpu_model_runner.py` (~line 998)

With `_get_mm_dummy_batch` no longer referencing `self.model_config_copy`, the `deepcopy` created at `__init__` may now be dead code allocating memory unnecessarily.

**Suggestion:** Check if `model_config_copy` is still used elsewhere. If not, either remove it or reuse it in the new helper (see Finding 2).

---

adobrzyn:
**Finding 4** 🟡 **Low-Medium** · `hpu_model_runner.py:L5219-L5251`

The old code raised `NotImplementedError` for unsupported modalities. The new `_get_dummy_mm_inputs_with_options` silently falls through to default options, which may produce confusing behavior if a new modality (e.g., audio) is added.

**Suggestion:** Add a log warning or explicit guard for unrecognized modalities.

---

adobrzyn (approved):

mgawarkiewicz-intel (approved):

PR #1376, opened 2026-04-21
fix: stabilize decode block_list growth for long-context workloads (v…

adobrzyn:
…0.19.0)

In non-contiguous PA mode, the decode block_list grows as num_seqs * blocks_per_seq which can exceed max_blocks * 3 for long-context workloads (e.g. 200K context). Each new unique block count triggers an HPU graph recompilation, causing a recompilation storm and OOM.

Fix 1 (hpu_model_runner.py): When block_bucket_size < len(block_list), round up via calc_fallback_value() for stable shapes instead of recompiling per unique block count.

Fix 2 (exponential.py): Raise the decode block cap using:
  blocks_per_seq = ceil(max_model_len / block_size)
  max(min(blocks_per_seq * max_num_seqs, max_blocks * max_num_seqs // 3),
      max_blocks * 3)
This covers realistic long-context scenarios while maintaining a bounded upper limit.

Validated on v0.19.0: 9 unique decode shapes (vs 1,996 baseline), P50 TPOT 242ms (vs 321ms), output throughput 155 tok/s (vs 112).

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -2083,17 +2083,14 @@ class HPUModelRunner(HpuKVConnectorModelRunnerMixin):
         block_bucket_size: int
         if self.use_contiguous_pa:
             actual_blocks_needed = max(block_list) + 1 if block_list else 0
-            actual_blocks_needed = min(actual_blocks_needed,
-                                       self._max_cache_blocks)
+            actual_blocks_needed = min(actual_blocks_needed, self._max_cache_blocks)
 
             block_bucket_size = \
                 self.bucketing_manager.find_decode_bucket(batch_size,
                                                           actual_blocks_needed)[2]
             if actual_blocks_needed > block_bucket_size:
                 block_bucket_size = min(
-                    calc_fallback_value(
-                        actual_blocks_needed,
-                        self.bucketing_manager.fallback_blocks_base_step),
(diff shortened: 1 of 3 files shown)


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -2101,6 +2101,9 @@ def padding_fn(tensor, pad_value):
                 self.bucketing_manager.find_decode_bucket(batch_size,
                                                           len(block_list))[2]
             block_bucket_size += self.get_dp_padding(block_bucket_size)
+            if block_bucket_size < len(block_list):
```
michalkuligowski, line 2109: get_dp_padding is normally how we guarantee all DP ranks end up with the same shape (it typically returns an all‑reduced max). The fallback branch replaces block_bucket_size with a value derived only from this rank's local len(block_list). Two ranks hitting this fallback with different local block-list lengths will produce different shapes → collective op shape mismatch / deadlock risk.

Recommended fix: after computing the fallback value, apply get_dp_padding again, e.g.

`
    
    if block_bucket_size < len(block_list):
        block_bucket_size = calc_fallback_value(
        len(block_list), self.bucketing_manager.fallback_blocks_base_step)
    
        block_bucket_size += self.get_dp_padding(block_bucket_size)`

adobrzyn: Should be fine now

kamil-kaczor (approved):
lgtm

mgawarkiewicz-intel (approved):

PR #1378, opened 2026-04-21
Set Docker auto calc PT_HPU_LAZY_MODE=0 as default

nngokhale:
now Torch compile default for ptupstream and ptfork 
add documentation


adobrzyn:
@nngokhale shouldn't we wait for CI to switch before merging it?

nngokhale:
@adobrzyn I have manually tested. Last release 0.17.1 torch compile was enabled as an option. Now decision with @PatrykWo is to make it default with lazy as optional. This way default entrypoint will work on both ptfork and ptupstream.

PatrykWo (approved):

PR #1384, opened 2026-04-22
Correct get_supported_kernel_block_sizes

jbyczkow:
Use explicit list instead of multiple of 16 condition.


vllm_gaudi/v1/attention/backends/hpu_attn.py
```diff
@@ -38,7 +38,7 @@ def get_supported_kernel_block_sizes() -> list[Union[int, MultipleOf]]:
         # 128 is the standard HPU kernel block size; 528 is required for
         # Granite 4.0-H (granitemoehybrid) without prefix caching (16-token
         # FA alignment), 768 with prefix caching (chunk-aligned).
-        return [MultipleOf(16)]
+        return [128, 528, 768]
```
michalkuligowski, line 41: fyi this is still a divisibility constraint, not exact match, allowing 256, 384 , 1536 as is resolved in https://github.com/vllm-project/vllm/blob/main/vllm/v1/attention/backend.py#L189

ksmusz (approved):

PR #1389, opened 2026-04-23
Bucketing edge cases finetune for longer ctx (#1362)

ksmusz:
### Problem

When using `VLLM_BUCKETING_FROM_FILE` with `max_model_len=131072` and a non-power-of-two `block_size` (e.g. 528 for Granite hybrid models), the `not_over_max_model_len` filter silently rejected the max context bucket (ctx=249). The old check `(query + ctx * block_size) <= max_model_len` failed because the padded bucket query (e.g. 8192) plus full context (249×528=131,472) exceeded 131,072 — even though at runtime the actual last-chunk query is shorter than the bucket size.

This caused two cascading failures:
1. Missing max-ctx bucket forced fallback bucket generation at runtime.
2. The fallback (`calc_fallback_value`) overshoots dramatically (e.g. ctx=235 → 448 blocks), triggering OOM during graph compilation.

Additionally, `mamba_decode_corrector` used `math.floor` to cap decode ctx, producing 1984 instead of the correct 1992 (`8 × ceil(131072/528)`). This caused repeated fallback warnings during decode.

### Changes

**`vllm_gaudi/extension/bucketing/common.py`**:
- Relaxed `not_over_max_model_len` to check `ctx <= ceil((max_model_len
- 1) / block_size)` — only context blocks matter, not the padded query size.
- Added prompt-specific fallback cap in `generate_fallback_bucket` to limit ctx to `ceil(max_model_len / block_size)`, preventing oversized fallback buckets.
- Fixed `mamba_decode_corrector` (`math.floor` → `math.ceil`) and removed it as it became identical to `correct_for_max_model_len`. Simplified `get_corrector` accordingly.

**`vllm_gaudi/v1/worker/hpu_model_runner.py`**:
- Fixed `_prepare_dummy_scenario` to cap warmup dummy total tokens when `query + ctx * block_size > max_model_len` in the non-merged-prefill path, recomputing context blocks so that `scheduled_tokens == prompt_query_len` exactly.

---------

### DIFF
--- a/vllm_gaudi/extension/bucketing/common.py
+++ b/vllm_gaudi/extension/bucketing/common.py
@@ -208,7 +208,7 @@ class HPUBucketingManager():
 
     ### RETRIEVE BUCKETS FUNCTIONS ###
 
-    def generate_fallback_bucket(self, batch_size, seq_len, ctx):
+    def generate_fallback_bucket(self, batch_size, seq_len, ctx, is_prompt=False):
         assert self.max_num_batched_tokens is not None
         new_batch_size = calc_fallback_value(batch_size, self.fallback_bs_base_step)
         if new_batch_size > self.max_num_seqs:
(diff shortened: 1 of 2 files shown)


jbyczkow (approved):

wpyszka (approved):
approved for 0.19.0

PR #1401, opened 2026-04-29
fix: HPU-specific bug fixes for KV-offload + async spec-decode (#1264)

hsubramony:
Bug 1 (hpu_async_scheduler): clamp num_external_computed_tokens to 0 in _update_requests_with_invalid_blocks() override. When OOM causes block invalidation the affected-token span can exceed the externally-computed prefix, incorrectly driving num_external_computed_tokens negative.

Bug 2 (hpu_async_scheduler): fix stale num_cached_tokens after preemption. After OOM preemption and requeue a request restarts from num_computed_tokens=0; the OffloadingConnector may assign new external cache hits leaving num_cached_tokens inconsistent (< num_external_computed_tokens). A schedule() post-processing pass detects and corrects this.

Bug 2b (utils): clamp PromptTokenStats.get_by_source() to 0 via monkey-patch. During the brief inconsistency window the Prometheus counter would crash with "Counters can only be incremented by non-negative amounts".

Bug 3 (hpu_model_runner): fix tensor shape mismatch [N,1] vs [N,M] in the async scheduling path of _create_decode_input_data when a spec-decode request has num_tokens > 1.

Bug 4 (hpu_model_runner): prevent Habana workspace OOM triggered by OffloadingConnector requeuing a decode request with many scheduled tokens. Route multi-token non-spec-decode requests through the prefill bucket path (which handles large context correctly) instead of the decode bucket path (which has no prepared bucket for batch_size=N*blocks, causing JIT recompile with a 107 GiB workspace allocation).

    Co-authored-by: GitHub Copilot

---------


afierka-intel (changes requested):
## Review: Request changes — one real blocker, the rest are polish

Harish — this is a fair port of #1264. Since the user asked me to re-check the three items I flagged on #1264, I lead with that, then note what the rebase onto `main` exposes.

### Status of the three #1264 comments

**1. Shape indexing in the async path (Bug 3) — addressed ✅**

`_create_decode_input_data` now slices `num_tokens_per_req[:num_decodes]` in `torch.split`, indexes the LHS as `token_ids_device[:num_decodes * num_tokens]`, and `.view(-1, 1)` on the RHS so both sides are `[num_decodes * num_tokens, 1]`. Matches the flattened CPU path (line ~2895). Good fix.

_Minor leftover:_ the CPU path at line 2865 still passes the full padded `num_tokens_per_req` (with trailing zeros) to `torch.split`, while the async path slices `[:num_decodes]`. Both work today but the asymmetry invites future bugs.

**2. `E402` lint on `utils.py` — addressed with `# noqa` ⚠️**

`# noqa: E402` is the pragmatic choice given the monkey-patch ordering constraint you explained on #1264. Please move that explanation **into the code** — a one-line comment next to the `# noqa` stating "cannot be hoisted: depends on `_async_sched_module.AsyncScheduler = HPUAsyncScheduler` above" would save the next maintainer from re-discovering it.

**3. Unit test for `ensure_multi_token_decodes_last` — still missing ❌**

On #1264 you said the test would land on main because verification needed more time. This PR *is* that main PR and still ships zero tests. At minimum I want one test that builds an `InputBatch` with:
- several 1-token decodes,
- one multi-token catch-up decode,
- one prompt,
- `speculative_config=None`,

and asserts the post-sort layout is `[1-tok decodes … | multi-tok decode | prompt]` and that `_get_prompts_and_decodes` routes the multi-token request to the prefill list. Given this is exactly the code path that produced the 107 GiB workspace OOM, I do not want to see it ship to main untested.

---

### New issues the port exposes

#### A. `ensure_decodes_first` diverges semantically from `_ensure_decodes_first` — **blocking**

The port adds a module-level `ensure_decodes_first(b: InputBatch)` at `hpu_model_runner.py:430` and stops calling the method `HPUModelRunner._ensure_decodes_first` (still present at line ~3936 as dead code). But the two are **not** equivalent:

- `_ensure_decodes_first` delegates to `self._is_prompt(i, scheduler_output)`, which considers:
  - `num_computed_tokens < num_prompt_tokens` (normal prompt),
  - `num_scheduled_tokens > num_decode_tokens` ("maybe preempted prompt", spec-decode aware),
  - `not self.is_decoder_only(req_id)`.
- The new `ensure_decodes_first` uses only `num_computed_tokens_cpu[i] < num_prompt_tokens[i]`.

Consequences:
1. A **preempted prompt** that has finished its prompt phase but re-schedules more than the spec-decode step count is classified as decode here but was prompt before — a behavior change for the exact OOM/requeue scenario this PR targets.
2. **Decoder-only** requests that were always treated as decode may now be classified as prompt when `num_computed < num_prompt`.
3. Spec-decode paths lose the scheduled-token check.

Please either (a) delete the old `_ensure_decodes_first` and replicate `_is_prompt`'s full logic in the module-level helper, or (b) keep the method form and add only `ensure_multi_token_decodes_last` as new. Having two helpers with subtly different semantics, one dead, is a trap.

#### B. `HPUAsyncScheduler.schedule()` post-processing only touches `self.running` — question

```python
for request in self.running:
    if request.num_cached_tokens < request.num_external_computed_tokens:
        request.num_cached_tokens = request.num_computed_tokens
```

A request requeued by `OffloadingConnector` that did **not** fit this step stays on `self.waiting` — the invariant violation you guard against can persist there until the next step. Is that intentional? If the Prometheus crash (Bug 2b) can fire from a waiting-queue path, the monkey-patch in `utils.py` is the only safety net.

Also, the docstring frames this as the "`num_computed_tokens=0`" case, but the assignment relies on `super().schedule()` having already advanced `num_computed_tokens`. Please clarify the docstring.

#### C. `_update_requests_with_invalid_blocks` is a full copy of upstream with one changed line — maintenance risk

~80 lines reproduced; the only functional delta is `max(0, …)` on `num_external_computed_tokens`. If upstream evolves this routine (hybrid memory allocator — the existing `TODO (davidb)` — new connector types) we silently diverge. Please:
- Add a `# Keep in sync with vllm.v1.core.sched.async_scheduler.AsyncScheduler._update_requests_with_invalid_blocks@<commit>` pointer.
- Highlight the single-line delta with an inline comment.
- File an upstream issue on `vllm-project/vllm` — I asked for this on #1264 as well and cannot find it linked here.

#### D. Assertion in the prompt-traversal loop — minor

```python
assert num_computed_tokens < num_prompt_tokens or (
    num_scheduled_tokens > 1 and not self.vllm_config.speculative_config), (...)
```

The predicate is "true prompt OR multi-token catch-up decode re-routed by `ensure_multi_token_decodes_last`". If it ever fires, the on-call engineer needs to know which limb failed and which request. Split into two asserts, or at least include `req_id` — `idx i` is not stable across reorders.

#### E. Style nits

- Two trailing blank lines after `ensure_multi_token_decodes_last` before the next `def`.
- `scheduled_tokens: dict` → prefer `Mapping[str, int]` / `dict[str, int]` for an explicit contract.
- Duplicate `num_scheduled_tokens = scheduler_output.num_scheduled_tokens[req_id]` inside the new assertion block (it is also computed a few lines below in the original code).

#### F. Merge hygiene — minor

The PR is `Merge branch 'main' into port_pr1264` on top of `0c722463`. A rebase would keep `main` linear; not a blocker if repo convention allows merge commits.

#### G. #1264 follow-ups still open

- Bug 2b: moving the Prometheus clamp out of module-level `utils.py` into the scheduler scope.
- Upstream vLLM report for Bug 1.
- Unit test (see item 3).

---

### Verdict

**Requesting changes** on point **A** only — the other items are polish that can land as a follow-up if the team wants to unblock this. Point A is a real behavior change for preempted prompts, spec-decode and decoder-only flows in the exact scenario the PR is trying to stabilize, and should not slip silently into main.

Thanks again for pushing this through. 🙏

iboiko-habana:
Please fix pre-commit issue https://github.com/vllm-project/vllm-gaudi/actions/runs/25307416414/job/74188974820

And enable pre-commit
pip install pre-commit
pre-commit install

afierka-intel (approved):

PR #1410, opened 2026-05-04
Clarify VLLM_PROMPT_BS_BUCKET_MAX runtime behavior in docs

iboiko-habana:


docs/configuration/env_variables.md
adobrzyn, line 27: I would have added this info here as well, something like that:

`
| `VLLM_PROMPT_BS_BUCKET_MAX` | Sets prefill batch size | `1` |
`

adobrzyn (approved):

PR #1419, opened 2026-05-06
Fix the FP8 KV Cache accuracy issue (Minimax-M2.5) by using B2BMatmul only

testdig:
This issue happens when input context length is greater than 20000 tokens

GSM8K on G2D
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9257|±  |0.0072|
|     |       |strict-match    |     5|exact_match|↑  |0.9204|±  |0.0075|


czhu15 (approved):
LGTM

yangulei (approved):
LGTM, nice catch!

PR #1431, opened 2026-05-09
Fix conv_state corruption in compact GDN batched decode

osavchenkox:
When HPU bucketing pads a decode batch (e.g., 5 real sequences padded to bucket size 8), prepare_mamba_state_idxs fills padding positions with _MAMBA_PAD_BLOCK_ID = num_blocks (e.g., 6947). With compact GDN enabled, state tensors have only compact_total slots (e.g., 26). Downstream code uses torch.remainder(index, num_slots) to map indices to valid slots. This causes remainder(6947, 26) = 5 — mapping padding to a real request's slot.

Padding tokens (with garbage embeddings from bucketing) then overwrite the real request's conv_state via conv_states[safe_cache_idx] = new_state. On the next decode step, the affected request reads corrupted conv_state, producing garbage q/k/v projections and nonsensical output.

The fix uses -1 as the padding value for compact GDN groups. torch.remainder(-1, 26) = 25 correctly maps to the last slot — a dedicated garbage slot allocated by the +2 in compact_total = max_reqs * num_groups + 2. For non-compact groups, _MAMBA_PAD_BLOCK_ID (= num_blocks) is preserved, as remainder(num_blocks, num_blocks + 1) = num_blocks (the last slot) works correctly.


iboiko-habana:
Please fix pre-commit https://github.com/vllm-project/vllm-gaudi/actions/runs/25588982990/job/75123027236

adobrzyn:
**Potential same issue in prefill prefix-caching path**

Lines 2545 and 2552 in `vllm_gaudi/v1/worker/hpu_model_runner.py` also use `_MAMBA_PAD_BLOCK_ID` (= `num_blocks`) as a fill value in `mamba_chunks_to_block_mapping_cpu` and `bcr_padded`, iterating over ALL groups including compact GDN ones:

```python
# Line 2545
mamba_chunks_to_block_mapping_cpu = torch.full((nphysical_chunks, ),
                                               self._MAMBA_PAD_BLOCK_ID,  # = num_blocks
                                               ...)

# Line 2552
bcr_padded = torch.full((max_cached_blocks, ),
                        self._MAMBA_PAD_BLOCK_ID,   # = num_blocks
                        ...)
```

In `granite_causal_conv1d.py` (line 105), `blocks_caching_range` (derived from `bcr_padded`) is used directly to index `conv_states`:
```python
conv_states[blocks_caching_range, -state_len:, :] = new_states.transpose(-1, -2)
```

If padded positions with value `num_blocks` are used to index a compact tensor (with only `compact_total` slots), this could cause either an OOB error or a similar corruption to the one fixed here.

Could you verify whether these padded positions can ever reach a `conv_states` indexing operation for compact GDN groups in the prefix-caching path?

osavchenkox:
@adobrzyn good catch — I traced the path and concluded it's **safe in practice, but only by virtue of model coverage, not by code-level gating**. Details:

1. The padded `bcr_padded` / `mamba_chunks_to_block_mapping_cpu` tensors at lines 2602/2595 are built per-group and stacked into `[num_groups, ...]`, but their **only consumer** is `granite_causal_conv1d_fn` / `_update` (see [granite_causal_conv1d.py docstring](https://github.com/vllm-project/vllm-gaudi/blob/dev/osavchenko/fix_compact_mode/vllm_gaudi/ops/granite_causal_conv1d.py#L13)): _"Used exclusively by hpu_mamba_mixer2.py (Granite 4.0). Other models continue to use causal_conv1d_pytorch.py."_
2. `hpu_mamba_mixer2` is **Mamba2-only** (Granite-4). GDN-attention / linear-attention layers (Qwen3.5-Next, MiniMax-M2) live in `vllm_gaudi/models/qwen3_5.py` and never call into this path — `blocks_caching_range` is wired through `MambaAttentionMetadata` but ignored by the GDN mixer.
3. Compact-GDN allocation only fires when `kv_cache_spec.mamba_type in _GDN_MAMBA_TYPES` (`gdn_attention`, `linear_attention`) — see hpu_model_runner.py line 6058. Granite-4 layers are `mamba2`, so `_compact_gdn_group_ids` is always empty for them.
4. There's no shipped hybrid that mixes Mamba2 **and** GDN in the same model, so the cross-product (compact GDN group + Granite prefix-caching consumer) doesn't materialize.

Net effect today: for compact-GDN groups in the prefix-caching path we **build** a `bcr_padded` filled with `_MAMBA_PAD_BLOCK_ID = num_blocks` and stack it, but no kernel ever indexes `conv_states` with it. So no OOB / no corruption — but the tensors are wasted work for those groups.

I'd treat the underlying defensive fix as a separate, smaller follow-up to keep this PR scoped to the decode `state_idxs` corruption. Two reasonable options for the follow-up:
- Skip building `bcr_padded` / `mamba_chunks_to_block_mapping_cpu` for `group_idx in self._compact_gdn_group_ids` (cheapest, removes the latent foot-gun).
- Or use `pad_val = -1` for compact-GDN groups here too, mirroring the decode fix — defensive against any future Mamba2+GDN hybrid that might wire these tensors into a compact `conv_states`.

WDYT — fold a small extra commit here, or split into a separate PR after this one merges?

adobrzyn:
@osavchenkox - sure, you can create different PR for that

iboiko-habana (approved):

PR #1436, opened 2026-05-11
[FIX_FOR_VLLM_CUSTOM=dcacdf9a8860a86401127d1c8f93ebf3cfbfd026] Fix MultiModelEngineClient, Qwen3.5 compilation, and EPLB refactoring

pawel-olejniczak:
Fix upstream regressions affecting hourly CI:

1. **MultiModelEngineClient**: Added missing `notify_kv_transfer_request_rejected` abstract method (upstream PR vllm-project/vllm#41269)
2. **Qwen3.5 test harness**: Updated `test_common.py` to read `enforce_eager` from model card config (with env var override), enabling per-model compilation control
3. **EPLB refactoring**: Removed `EMPTY_EPLB_STATE` import and `enable_eplb` parameter from `patched_create_fused_moe_router` after upstream MoE refactor (upstream PR vllm-project/vllm#41055)

Note: The `enforce_eager: true` workaround for Qwen3.5 compilation has been removed — the root cause (mamba_type str-vs-Enum comparison in hybrid cache allocation) is properly fixed by #1449, which should merge first.

Verified on HPU: unit tests pass on Gaudi 3 (MoE, FP8, compressed tensors).


tests/full_tests/model_cards/qwen3.5-35b-a3b.yaml
```diff
@@ -12,6 +12,7 @@ model_card:
   reasoning_parser: qwen3
   max_num_batched_tokens: 16384
   max_gen_toks: 2048
+  enforce_eager: true
```
shepark: The compilation failure (aot_autograd() does not yet handle input mutations on views with different dtypes) is not inherent to the Qwen3.5 model — it's caused by a type mismatch in the hybrid cache allocation logic in hpu_model_runner.py.

Root Cause:
Upstream vllm commit [5536fc0c0] changed MambaSpec.mamba_type from str to MambaAttentionBackendEnum. However, the hybrid cache allocation code in hpu_model_runner.py still compares mamba_type against string literals ("gdn_attention", "linear_attention"). Since Enum != str, these comparisons silently evaluate to False, causing GDN layers to fall through to the standard Mamba2 shared-buffer path. That path creates conv_state (bf16) and ssm_state (fp32) as views on the same raw buffer, which triggers the aot_autograd assertion when PyTorch detects aliased inputs with different dtypes being mutated.

So, I prepared PR for that. (https://github.com/vllm-project/vllm-gaudi/pull/1449)

tests/full_tests/model_cards/qwen3.5-35b-a3b.yaml
```diff
@@ -12,6 +12,7 @@ model_card:
...
```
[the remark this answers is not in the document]
pawel-olejniczak: I've updated test_common.py to read enforce_eager from the model config (with env var override), so the field works correctly now. However, I've removed enforce_eager: true from the Qwen3.5 model card itself (commit 8ecd3b9e) because it's a performance-degrading workaround. The root cause (mamba_type str-vs-Enum comparison) is properly fixed in #1449.

tests/full_tests/model_cards/qwen3.5-35b-a3b.yaml
pawel-olejniczak: Thanks for the root cause analysis and the proper fix in #1449! I've removed the enforce_eager: true from this PR (commit 8ecd3b9e). Your PR #1449 should merge first since it runs against the stable vLLM commit and won't hit the EPLB import error. Once #1449 is in, this PR (#1436) will advance the stable commit and the Qwen3.5 test should pass without the workaround.

iboiko-habana (approved):

PR #1438, opened 2026-05-12
Fix accuracy issue for the sliced FusedSDPA

yangulei:
- Fix the accuracy issue for the INC path when slicing is enabled for FusedSDPA.
- Fix the accuracy drop for the `is_causal=True` chunks by always passing `attn_mask` for causal chunks. The FusedSDPA kernel use different strategy internally which cause accuracy drop for the returned `m` and `linv`.


czhu15 (approved):
LGTM

czhu15 (approved):
LGTM

PR #1439, opened 2026-05-12
Fix patch_hf3fs_mock_client_for_cpu_only

hsubramony:
1) added in https://github.com/vllm-project/vllm-gaudi/pull/1453
16 is supported for testing/smaller models; 128 is the standard HPU
kernel block size; 528 is required for Granite 4.0-H
(granitemoehybrid) without prefix caching (16-token FA alignment),
768 with prefix caching (chunk-aligned).

2) _patch_hf3fs_mock_client_for_cpu_only
Upstream mock client unconditionally calls
    ``torch.cuda.current_stream().wait_event(event)`` in ``batch_write``.
    In environments where PyTorch is not compiled with CUDA, that path throws
    and the method returns ``-1`` for writes, causing connector unit tests to
    fail. This patch keeps the same behavior but skips CUDA synchronization when
    CUDA is unavailable.


kamil-kaczor (approved):
lgtm

iboiko-habana (approved):

PR #1447, opened 2026-05-14
Fix decode bucket filter issues from #1122

yangulei:
## Fixes

Two bugs introduced by #1122 (commit f24f3f9d):

### 1. IndexError when using file-based bucketing (GAUDISW-248587)
When `VLLM_BUCKETING_FROM_FILE` is used (e.g. GraniteMoeHybrid model), `ctx_range` is passed as an empty list to `generate_buckets()`. The `num_ctx_tokens_less_or_equal_batched_max_model_len` filter accessed `ctx_range[0]` unconditionally, causing `IndexError: list index out of range`.

**Fix**: Safe access with fallback to 0 when `ctx_range` is empty.

### 2. Contiguous PA decode buckets incorrectly filtered (GAUDISW-248598)
The ctx filter was applied to contiguous PA decode buckets, incorrectly dropping valid buckets. For example, with `max_model_len=2048`, `block_size=256`, `max_num_seqs=256`, bucket `(256, 1, 2112)` was filtered because `2112 > ceil(2048/256)*256 = 2048`, but 2112 is a valid user-configured `VLLM_DECODE_BLOCK_BUCKET_MAX`.

**Fix**: Remove the ctx filter from contiguous PA decode buckets. For contiguous PA, the block range is already bounded by `max_blocks` in the bucketing strategies.

## Tests
- Added `test_file_buckets_with_empty_ctx_range_no_crash` — reproduces the server.log IndexError
- Added `test_contiguous_pa_decode_buckets_not_filtered_by_ctx` — reproduces the std_out.txt issue
- Narrowed `test_decode_buckets_satisfy_ctx_filter` to non-contiguous PA only
- Updated docstrings

### DIFF
--- a/tests/unit_tests/test_bucketing.py
+++ b/tests/unit_tests/test_bucketing.py
@@ -771,25 +771,27 @@ def test_contiguous_pa_decode_buckets_not_filtered_by_ctx(monkeypatch):
     assert max_blocks in bucket_ctxs, (f"Bucket ctx={max_blocks} (num_hpu_blocks) was incorrectly filtered out.")
 
 
-def test_file_buckets_with_empty_ctx_range_no_crash(monkeypatch):
-    """When VLLM_BUCKETING_FROM_FILE is used, ctx_range is empty.
+def test_file_buckets_bypass_filters(monkeypatch):
+    """File-based bucketing (VLLM_BUCKETING_FROM_FILE) skips all filters.
 
-    The ctx filter must not crash with IndexError on ctx_range[0].
-    Reproduces server.log issue: GraniteMoeHybrid model with file-based
-    bucketing caused IndexError in num_ctx_tokens_less_or_equal_batched_max_model_len.
+    Buckets (1,1,256) and (2,1,512) would normally be rejected by the
+    batch_size_smaller_than_blocks or ctx filters in non-file mode.
+    Since file buckets bypass filters entirely, all provided buckets
+    must appear in the output unchanged.
     """
-    monkeypatch.setenv("VLLM_CONTIGUOUS_PA", "false")
+    monkeypatch.setenv("VLLM_CONTIGUOUS_PA", "true")
     clear_config()
     get_config()
 
-    max_model_len = 131072
-    block_size = 528
+    max_model_len = 2048
+    block_size = 256
     max_num_seqs = 32
     max_blocks = 2424
 
-    file_buckets = [(1, 1, 256), (1, 1, 512), (2, 1, 256), (2, 1, 512), (32, 1, 2424)]
+    # (512,1,256) would be rejected by batch_size_smaller_than_blocks (bs > ctx)
+    # All buckets pass through because file_buckets bypass filters entirely
+    file_buckets = [(1, 1, 256), (1, 1, 512), (2, 1, 256), (512, 1, 256), (32, 1, 2424)]
 
-    # ctx_range is empty when using file-based bucketing
     buckets = generate_buckets(
         bs_range=[],
         query_range=[],
@@ -804,4 +806,5 @@ def test_file_buckets_with_empty_ctx_range_no_crash(monkeypatch):
         file_buckets=file_buckets,
     )
 
-    assert len(buckets) > 0, "Should produce buckets from file_buckets"
+    assert set(buckets) == set(file_buckets), (f"All file buckets should pass through unfiltered.\n"
+                                               f"Expected: {sorted(file_buckets)}\nGot: {sorted(buckets)}")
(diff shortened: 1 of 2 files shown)


yangulei:
@iboiko-habana @kamil-kaczor 
Please help to review this PR. It has been validated by the reporters for GAUDISW-248598 and GAUDISW-248587.
Thanks!

adobrzyn:
**Finding 1** 🔴 **Critical** · `vllm_gaudi/extension/bucketing/common.py:L451`

With `ctx_min = ctx_range[0] if ctx_range else 0`, the file-bucketing path (`ctx_range == []`) sets `ctx_min = 0`, so the condition `ctx > ctx_min` is true for every bucket with `ctx > 0` and the filter is now applied to user-provided file buckets — silently rejecting any whose `ctx > ceil(max_model_len/block_size) * bs`.

Worked example using parameters from the new `test_file_buckets_with_empty_ctx_range_no_crash` (`max_model_len=131072`, `block_size=528`, non-contiguous PA, `file_buckets=[(1,1,256), (1,1,512), (2,1,256), (2,1,512), (32,1,2424)]`):

- `ceil(131072/528) = 249`
- `(1,1,256)`: `1*249=249`, `256 > 249` → **dropped**
- `(1,1,512)`: `512 > 249` → **dropped**
- `(2,1,256)`: `2*249=498`, `256 ≤ 498` → kept
- `(2,1,512)`: `512 > 498` → **dropped**
- `(32,1,2424)`: `32*249=7968`, `2424 ≤ 7968` → kept

3 of 5 explicitly-configured file buckets are silently filtered out. The new test only asserts `len(buckets) > 0`, which masks this. The `num_ctx_tokens_less_or_equal_batched_max_model_len` filter from #1122 was designed to limit auto-generated warmup explosion — it should not be silently rejecting buckets the user explicitly configured via `VLLM_BUCKETING_FROM_FILE`.

**Suggestion:** Skip filters entirely for file-provided buckets (file = explicit user intent), e.g. in `generate_buckets`:

```python
filters = [] if file_buckets else get_filters(is_prompt, use_merged_prefill, use_contiguous_pa)
```

This preserves #1122's filter for auto-generated buckets (its actual target) and avoids any behavior change on the file path beyond "don't crash". Also strengthen the regression test to assert all five file buckets are preserved:

```python
assert len(buckets) == len(file_buckets)
for fb in file_buckets:
    assert fb in buckets
```

---

yangulei:
> **Finding 1** 🔴 **Critical** · `vllm_gaudi/extension/bucketing/common.py:L451`
> 
> With `ctx_min = ctx_range[0] if ctx_range else 0`, the file-bucketing path (`ctx_range == []`) sets `ctx_min = 0`, so the condition `ctx > ctx_min` is true for every bucket with `ctx > 0` and the filter is now applied to user-provided file buckets — silently rejecting any whose `ctx > ceil(max_model_len/block_size) * bs`.
> 
> Worked example using parameters from the new `test_file_buckets_with_empty_ctx_range_no_crash` (`max_model_len=131072`, `block_size=528`, non-contiguous PA, `file_buckets=[(1,1,256), (1,1,512), (2,1,256), (2,1,512), (32,1,2424)]`):
> 
> * `ceil(131072/528) = 249`
> * `(1,1,256)`: `1*249=249`, `256 > 249` → **dropped**
> * `(1,1,512)`: `512 > 249` → **dropped**
> * `(2,1,256)`: `2*249=498`, `256 ≤ 498` → kept
> * `(2,1,512)`: `512 > 498` → **dropped**
> * `(32,1,2424)`: `32*249=7968`, `2424 ≤ 7968` → kept
> 
> 3 of 5 explicitly-configured file buckets are silently filtered out. The new test only asserts `len(buckets) > 0`, which masks this. The `num_ctx_tokens_less_or_equal_batched_max_model_len` filter from #1122 was designed to limit auto-generated warmup explosion — it should not be silently rejecting buckets the user explicitly configured via `VLLM_BUCKETING_FROM_FILE`.
> 
> **Suggestion:** Skip filters entirely for file-provided buckets (file = explicit user intent), e.g. in `generate_buckets`:
> 
> ```python
> filters = [] if file_buckets else get_filters(is_prompt, use_merged_prefill, use_contiguous_pa)
> ```
> 
> This preserves #1122's filter for auto-generated buckets (its actual target) and avoids any behavior change on the file path beyond "don't crash". Also strengthen the regression test to assert all five file buckets are preserved:
> 
> ```python
> assert len(buckets) == len(file_buckets)
> for fb in file_buckets:
>     assert fb in buckets
> ```
> 
> _[- Reviewed by Awesome ChlOpus]_

Done by [Skip filters for file-based bucketing and refactor tests](https://github.com/vllm-project/vllm-gaudi/pull/1447/commits/77b74a90ce8a513a16004e2999e2b9f92694fdb6).

kamil-kaczor (approved):
lgtm

PR #1449, opened 2026-05-14
Fix mamba_type comparison for GDN hybrid cache allococation

shepark:
Upstream vllm commit 5536fc0c0 changed MambaSpec.mamba_type from str to MambaAttentionBackendEnum. The hybrid cache allocation in hpu_model_runner.py still compared against str literals, causing GDN layers to fall through to the Mamba2 shared-buffer path. This created mixed-dtype views (bf16 conv_state+fp32 ssm_state) on the same storage, triggering an aot_autograd assertion error during compilation.

Use a module-level _GDN_MAMBA_TYPES tuple that includes both enum values and string literals for backward compatibility with older upstream versions.

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -50,6 +50,12 @@ from vllm.model_executor.layers.attention_layer_base import AttentionLayerBase
 from vllm.model_executor.layers.fused_moe.layer import FusedMoE
 from vllm.model_executor.layers.layernorm import RMSNorm
 from vllm.model_executor.layers.mamba.abstract import MambaBase
+try:
+    from vllm.v1.attention.backends.registry import MambaAttentionBackendEnum
+    _GDN_MAMBA_TYPES = (MambaAttentionBackendEnum.GDN_ATTN, MambaAttentionBackendEnum.LINEAR, "gdn_attention",
+                        "linear_attention")
+except (ImportError, AttributeError):
+    _GDN_MAMBA_TYPES = ("gdn_attention", "linear_attention")
 from vllm.model_executor.layers.vocab_parallel_embedding import (VocabParallelEmbedding)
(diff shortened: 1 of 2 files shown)


adobrzyn:
**Finding: `hpu_worker.py` has the same broken string-literal comparisons (not fixed)**

This PR only patches `hpu_model_runner.py`, but `vllm_gaudi/v1/worker/hpu_worker.py` has **4 more instances** of the identical `mamba_type in ("gdn_attention", "linear_attention")` pattern (lines 445, 448, 466, 488). These will break in exactly the same way after the upstream enum change.

Particularly dangerous are the `not in` checks (lines 448, 488) — with the enum change, GDN/linear layers will erroneously match the "standard Mamba2" path because `MambaAttentionBackendEnum.GDN_ATTN not in ("gdn_attention", "linear_attention")` evaluates to `True`. This could lead to incorrect memory calculations for hybrid models.

Suggestion: apply the same `_GDN_MAMBA_TYPES` pattern (or import it from a shared location) in `hpu_worker.py` as well.

adobrzyn:
**Finding 1** 🔴 **Critical** · `vllm_gaudi/v1/worker/hpu_worker.py:L445-L488`

The PR fixes string-literal `mamba_type` comparisons in `hpu_model_runner.py` but **misses the same pattern in `hpu_worker.py`** (lines 445, 448, 466, 488). These will break identically after the upstream enum change.

Particularly dangerous are the `not in` checks (lines 448, 488) — with the enum change, `MambaAttentionBackendEnum.GDN_ATTN not in ("gdn_attention", "linear_attention")` evaluates to `True`, causing GDN/linear layers to erroneously match the "standard Mamba2" path. This could lead to incorrect memory calculations for hybrid models.

**Suggestion:** Apply the same `_GDN_MAMBA_TYPES` pattern to `hpu_worker.py` — either duplicate the guarded import + tuple definition, or extract it into a shared location both files can import.

---

iboiko-habana (approved):

PR #1456, opened 2026-05-19
fix unit_moe: use loaded weight scales, force activation scale=1.0

yangulei:
The previous unit_moe implementation forced both weight and activation scales to 1.0, which caused accuracy issues in some cases.

This fix changes the behavior to:
- **Use the loaded scales for weights** (no longer force to 1.0)
- **Force scale=1.0 only for activation** (both dynamic and static paths)

Changes:
- Removed the weight rescaling block in `fp8_channel_moe_prepare_weights` that was absorbing scales into weights
- In `VllmMixtureOfExpertsOpFP8PerChannel.forward`: weight scales always use loaded values; only activation scales (`x_scale`, `w2_input_scale`) are forced to 1.0 when `enable_unit_moe` is set

### DIFF
--- a/vllm_gaudi/extension/ops.py
+++ b/vllm_gaudi/extension/ops.py
@@ -1008,16 +1008,6 @@ def fp8_block_moe_prepare_weights(layer, force_channel_fp8=False):
 
 
 def fp8_channel_moe_prepare_weights(layer):
-    if get_config().enable_unit_moe:
-        w13_weight = layer.w13_weight.to(torch.float)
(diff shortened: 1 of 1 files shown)


czhu15:
will this change impacts the runtime performance?

yangulei:
> will this change impacts the runtime performance?

Not sure.
This PR is for the accuracy issue of AA_LCR benchmark reported by Yinghao. He will check the performance and accuracy of this change. I will change it to 'Ready for review' if the performance and accuracy both ok.

PR #1457, opened 2026-05-19
[DO_NOT_MERGE] Use l2Norm TPC kernel for HPU compile

osavchenkox:
Replace eager _l2norm_last_dim with TPC l2_norm op, remove @torch._dynamo.disable to allow torch.compile, and fix dtype handling so l2norm outputs fp32 directly.


osavchenkox:
Rebased on latest `main` and dropped the `[DO_NOT_MERGE]` tag. The previous CI failure was an old workflow-version mismatch and not a code issue. Ready for review and pre-merge approval.

PR #1485, opened 2026-05-24
Fix decode bucket generation for hybrid models with mismatched block sizes

yangulei:
## Problem

For hybrid models like Qwen3.5 (GDN + attention), `_align_hybrid_block_size()` sets `block_size=640` (unified KV-cache page for mamba/attention alignment), while HPU kernels use `attn_block_size=128`.

The decode bucket generation (introduced by f24f3f9d) uses the formula:
```
max_decode_blocks = ceil(max_model_len / block_size) * max_num_seqs
                  = ceil(262144 / 640) * 45 = 18450
```

But the runtime decode path (`_create_decode_input_data`) computes `num_blocks` using `attn_block_size=128`, producing values up to `ceil(262144/128) * 45 = 92160`.

This causes hundreds of **"Configuration was not warmed-up"** warnings and costly HPU graph recompilation on every decode step.

## Root Cause

Two different block_size semantics coexist:
- `self.block_size = 640`: KV-cache management page size (unified for hybrid mamba/attention)
- `self.attn_block_size = 128`: HPU attention kernel page size (what hardware actually uses)

Decode bucket generation used `block_size` but should use `attn_block_size` to match the runtime.

## Fix

Temporarily scope `bucketing_manager.block_size` to `attn_block_size` during decode bucket generation in `warmup_model()`, then restore the original value so prompt fallback paths remain unaffected.

## Testing

- Verified with Qwen3.5-35B-A3B on 4x Gaudi3 (TP=4, max_model_len=262144, max_num_seqs=45)
- Decode buckets now correctly cover runtime num_blocks range
- No more "Configuration was not warmed-up" warnings during serving

### DIFF
--- a/tests/unit_tests/test_decode_bucket_hybrid.py
+++ b/tests/unit_tests/test_decode_bucket_hybrid.py
@@ -1,0 +1,440 @@
+# SPDX-License-Identifier: Apache-2.0
+###############################################################################
+# Copyright (C) 2024-2026 Intel Corporation
+#
+# This source code is licensed under the Apache 2.0 license found in the
+# LICENSE file in the root directory of this source tree.
+###############################################################################
+"""
+Regression tests for decode bucket generation and warmup in hybrid models.
+
+Hybrid models (e.g., Qwen3.5) have block_size != attn_block_size:
+- block_size=640: unified page size for KV cache management
+- attn_block_size=128: HPU kernel page size used by paged attention
+
+The decode path (_create_decode_input_data) computes num_blocks using
+attn_block_size. Therefore:
+1. Decode buckets MUST be generated in attn_block_size units.
+2. Warmup seq_lengths MUST produce the correct sum(num_blocks) to match
+   the target bucket after find_decode_bucket lookup.
+3. For non-contiguous PA, _generate_seq_lengths MUST NOT cap num_blocks
+   at kv_cache_config.num_blocks (physical pool), because runtime can
+   exceed this via prefix-sharing.
+
+Regression: f24f3f9d introduced a formula for max_decode_blocks using
+block_size instead of attn_block_size, and added a physical-pool cap
+that prevented large decode buckets from being warmed.
+"""
+
+import math
+
+import pytest
+from types import SimpleNamespace
+from unittest.mock import patch
+
+from vllm_gaudi.extension.bucketing.common import (
+    HPUBucketingManager,
+    find_equal_or_closest_greater_config,
+)
+from vllm_gaudi.extension.bucketing.exponential import ExponentialBucketingStrategy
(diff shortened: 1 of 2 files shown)


tests/unit_tests/test_decode_bucket_hybrid.py
```diff
@@ -0,0 +1,456 @@
+# SPDX-License-Identifier: Apache-2.0
+###############################################################################
+# Copyright (C) 2024-2025 Intel Corporation
```
kamil-kaczor: update the date

tests/unit_tests/test_decode_bucket_hybrid.py
```diff
@@ -0,0 +1,456 @@
...
+    """Minimal mock of HPUModelRunner for _generate_seq_lengths testing."""
+
+    def __init__(self, use_contiguous_pa, num_blocks, max_model_len, speculative_config=None):
+        self.use_contiguous_pa = use_contiguous_pa
+        self.kv_cache_config = SimpleNamespace(num_blocks=num_blocks)
+        self.max_model_len = max_model_len
+        self.speculative_config = speculative_config
+
+
+def _generate_seq_lengths(runner, num_samples, num_blocks, block_size):
```
kamil-kaczor, line 124: Please import his function from model runner instead of implementing it again. If it changes we won't catch an issue

kamil-kaczor (changes requested):

adobrzyn:
**Finding 1** 🟡 **Medium** · `tests/unit_tests/test_decode_bucket_hybrid.py:L122-141`

The test file defines a local `_generate_seq_lengths()` that reimplements the production method instead of importing/patching the real `HPUModelRunner._generate_seq_lengths`. As written, the tests will pass even if the production method in `vllm_gaudi/v1/worker/hpu_model_runner.py` regresses (e.g., someone re-introduces the unconditional cap). The same risk applies to `_MockModelRunner` — it only encodes the assumed invariants, not the real ones.

**Suggestion:** Either (a) instantiate the real `HPUModelRunner` with mocks and call `runner._generate_seq_lengths(...)`, or (b) `import` the function (or extract it into a pure helper shared by production and tests). Otherwise this file behaves more like documentation than a regression test.

---

tests/unit_tests/test_decode_bucket_hybrid.py
yangulei: done

tests/unit_tests/test_decode_bucket_hybrid.py
yangulei: done

kamil-kaczor (approved):
lgtm

PR #1486, opened 2026-05-24
[v0.21.0] Fix decode bucket generation for hybrid models with mismatched block sizes

yangulei:
## Problem

Backport of #1485 to releases/v0.21.0.

For hybrid models like Qwen3.5 (GDN + attention), `_align_hybrid_block_size()` sets `block_size=640` (unified KV-cache page), while HPU kernels use `attn_block_size=128`.

Decode bucket generation uses the formula:
```
max_decode_blocks = ceil(max_model_len / block_size) * max_num_seqs
                  = ceil(262144 / 640) * 45 = 18450
```

But the runtime decode path computes `num_blocks` using `attn_block_size=128`, producing values up to `92160`, causing hundreds of **"Configuration was not warmed-up"** warnings and HPU graph recompilation.

## Fix

1. Temporarily scope `bucketing_manager.block_size` to `attn_block_size` during decode bucket generation in `warmup_model()`, then restore.
2. Use `attn_block_size` in `_prepare_dummy_scenario()` for decode dummy data so warmup shapes match the generated buckets.

## Testing

- Verified with Qwen3.5-35B-A3B on 4x Gaudi3 (TP=4, max_model_len=262144, max_num_seqs=45)
- No more "Configuration was not warmed-up" warnings during serving

Fixes regression introduced by f24f3f9d.

### DIFF
--- a/tests/unit_tests/test_decode_bucket_hybrid.py
+++ b/tests/unit_tests/test_decode_bucket_hybrid.py
@@ -1,0 +1,440 @@
+# SPDX-License-Identifier: Apache-2.0
+###############################################################################
+# Copyright (C) 2024-2026 Intel Corporation
+#
+# This source code is licensed under the Apache 2.0 license found in the
+# LICENSE file in the root directory of this source tree.
+###############################################################################
+"""
+Regression tests for decode bucket generation and warmup in hybrid models.
+
+Hybrid models (e.g., Qwen3.5) have block_size != attn_block_size:
+- block_size=640: unified page size for KV cache management
+- attn_block_size=128: HPU kernel page size used by paged attention
+
+The decode path (_create_decode_input_data) computes num_blocks using
+attn_block_size. Therefore:
+1. Decode buckets MUST be generated in attn_block_size units.
+2. Warmup seq_lengths MUST produce the correct sum(num_blocks) to match
+   the target bucket after find_decode_bucket lookup.
+3. For non-contiguous PA, _generate_seq_lengths MUST NOT cap num_blocks
+   at kv_cache_config.num_blocks (physical pool), because runtime can
+   exceed this via prefix-sharing.
+
+Regression: f24f3f9d introduced a formula for max_decode_blocks using
+block_size instead of attn_block_size, and added a physical-pool cap
+that prevented large decode buckets from being warmed.
+"""
+
+import math
+
+import pytest
+from types import SimpleNamespace
+from unittest.mock import patch
+
+from vllm_gaudi.extension.bucketing.common import (
+    HPUBucketingManager,
+    find_equal_or_closest_greater_config,
+)
+from vllm_gaudi.extension.bucketing.exponential import ExponentialBucketingStrategy
+from vllm_gaudi.extension.runtime import get_config, clear_config
+from vllm_gaudi.v1.worker.hpu_model_runner import HPUModelRunner
+
+# --- Qwen3.5 hybrid model parameters ---
+_QWEN35_BLOCK_SIZE = 640  # unified page size (5 * 128)
+_QWEN35_ATTN_BLOCK_SIZE = 128  # HPU kernel page size
+_QWEN35_MAX_MODEL_LEN = 262144
+_QWEN35_MAX_NUM_SEQS = 45
+_QWEN35_NUM_HPU_BLOCKS = 15405  # physical blocks in attn_block_size units
+
+
+@pytest.fixture(autouse=True)
+def default_config(monkeypatch):
+    """Reset singleton and pin bucketing config for deterministic tests."""
+    # Reset singleton to prevent state leakage between tests
+    HPUBucketingManager._instance = None
+    # Pin bucketing strategy to avoid env-dependent behavior in CI
+    monkeypatch.setenv("VLLM_BUCKETING_STRATEGY", "exp")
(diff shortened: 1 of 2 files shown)


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -5288,11 +5292,19 @@ def _prepare_dummy_scenario(self, prompt_cfg, decode_cfg):
...
+            # Use attn_block_size for decode dummy data because decode buckets
+            # are generated in attn_block_size units (HPU kernel page size).
+            decode_blk_size = (self.attn_block_size if self.attn_block_size != self.block_size else self.block_size)
+            # Ensure at least 1 block per seq — some decode buckets may have
+            # ctx < batch_size (e.g., contiguous PA without bs <= ctx filter).
+            effective_num_blocks = max(decode_num_blocks, decode_bs)
             if self.use_contiguous_pa:
-                decode_seq_lengths = [self.block_size] * decode_bs
-                block_id = decode_num_blocks - 1
+                decode_seq_lengths = self._generate_seq_lengths(decode_bs, effective_num_blocks, decode_blk_size)
```
kamil-kaczor: Why lines around here differs from the PR to main? On main we call static list, here we call a function etc. Please verify you've CP from correct commit. Probably something like this will be correct, please check:
`decode_block_size = self.attn_block_size
if self.use_contiguous_pa:
    decode_seq_lengths = [decode_block_size] * decode_bs
    block_id = min(decode_num_blocks - 1, self.kv_cache_config.num_blocks - 1)
else:
    decode_seq_lengths = self._generate_seq_lengths(decode_bs, decode_num_blocks, decode_block_size)
    block_id = 0`

tests/unit_tests/test_decode_bucket_hybrid.py
```diff
@@ -0,0 +1,454 @@
+# SPDX-License-Identifier: Apache-2.0
+###############################################################################
+# Copyright (C) 2024-2025 Intel Corporation
```
kamil-kaczor: fix the date

tests/unit_tests/test_decode_bucket_hybrid.py
```diff
@@ -0,0 +1,454 @@
...
+    """Minimal mock of HPUModelRunner for _generate_seq_lengths testing."""
+
+    def __init__(self, use_contiguous_pa, num_blocks, max_model_len, speculative_config=None):
+        self.use_contiguous_pa = use_contiguous_pa
+        self.kv_cache_config = SimpleNamespace(num_blocks=num_blocks)
+        self.max_model_len = max_model_len
+        self.speculative_config = speculative_config
+
+
+def _generate_seq_lengths(runner, num_samples, num_blocks, block_size):
```
kamil-kaczor, line 124: import instead of rewriting it

kamil-kaczor (changes requested):

vllm_gaudi/v1/worker/hpu_model_runner.py
iboiko-habana: Not all PRs are ported from main to v0.21.0. Let's wait while baseline will be the same

adobrzyn:
**Finding 1** 🟡 **Medium** · `tests/unit_tests/test_decode_bucket_hybrid.py:L122-141`

The test file defines a local `_generate_seq_lengths()` that reimplements the production method instead of importing/patching the real `HPUModelRunner._generate_seq_lengths`. As written, the tests will pass even if the production method in `vllm_gaudi/v1/worker/hpu_model_runner.py` regresses (e.g., someone re-introduces the unconditional cap). The same risk applies to `_MockModelRunner` — it only encodes the assumed invariants, not the real ones.

**Suggestion:** Either (a) instantiate the real `HPUModelRunner` with mocks and call `runner._generate_seq_lengths(...)`, or (b) `import` the function (or extract it into a pure helper shared by production and tests). Otherwise this file behaves more like documentation than a regression test.

---

vllm_gaudi/v1/worker/hpu_model_runner.py
yangulei: Makes sense. Converting this PR back to draft. Will rebase and update once the v0.21.0 baseline aligns with main.

yangulei: Fixed — cherry-picked the correct commit from main and aligned the code. Now uses the same pattern: static list for contiguous PA, `_generate_seq_lengths` for non-contiguous, with `block_id = min(...)` cap.

tests/unit_tests/test_decode_bucket_hybrid.py
yangulei: Fixed — updated to 2024-2026.

tests/unit_tests/test_decode_bucket_hybrid.py
yangulei: Fixed — now imports `HPUModelRunner._generate_seq_lengths` via unbound method call instead of reimplementing.

yangulei:
@kamil-kaczor @adobrzyn @iboiko-habana 
All the previous comments are addressed. The CI is canceled while I tested this PR locally and the issues in GAUDISW-248817 and GAUDISW-248575 are fixed.
Please help to review again.

kamil-kaczor (approved):
lgtm

PR #1487, opened 2026-05-24
[DO_NOT_MERGE] Skip gather/index_copy for sequential state indices in GDN decode

osavchenkox:
When ssm_state_indices is None (sequential [0..N-1] case), work directly with final_state[:num_seqs] slice instead of gather + index_copy.

The sequential case applies when the caller knows state slots are contiguous from index 0, which is the common path in stable decode batches without preemption.


osavchenkox:
Rebased on latest `main`, marked ready-for-review, and dropped the `[DO_NOT_MERGE]` tag. Optimisation only fires when `ssm_state_indices` is None (sequential [0..N-1] case in stable decode batches without preemption); fall-through path is unchanged. Ready for pre-merge approval.

PR #1492, opened 2026-05-25
Port of: Fix stale gate ref overriding caller router_logits in dp_size==1 MoE fast path (#1469)

iboiko-habana:


kamil-kaczor (approved):
I trust you although Copilot says it's not needed. GL

PR #1498, opened 2026-05-26
Fix gpt-oss mxfp4 TP partitioning and quant_method matching

mkrze:
- Use cdiv-based per-rank partitioning for FusedMoE bf16 param layout, so non-block-aligned intermediate sizes (e.g. gpt-oss-120b: 2880 with tp_size=4) no longer mis-size the rank slice.
- For w2 (K = intermediate dim) dequantize over a block-aligned K window and crop back to the rank's true range via k_offset.
- Accept both 'mxfp4' and 'gpt_oss_mxfp4' as quant_method, since newer vLLM normalizes the checkpoint's 'mxfp4' to 'gpt_oss_mxfp4' in GptOssForCausalLMConfig.verify_and_update_model_config().


wpyszka (approved):
merging to include in nightly, isolated change

PR #1512, opened 2026-06-01
dynamic_quant: 20% improvements in per-channel mode

osavchenkox:
Replace max(dim=-1).values + unsqueeze with amax(dim=-1, keepdim=True). aten.max.dim builds a discarded indices tensor; amax skips it, and keepdim=True folds the unsqueeze into the reduce kernel.

Add standalone bench that compares bf16/per-tensor/per-channel/static FP8 with eager + torch.compile traces and FX graph dumps.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -199,3 +199,6 @@ shellcheck*/
 
 # Ingore moe/marlin_moe gen code
 csrc/moe/marlin_moe_wna16/kernel_*
+
+# Profiler / FX graph dumps from local bench scripts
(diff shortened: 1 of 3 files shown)


osavchenkox:
Friendly bump — this PR is ready for review. It targets per-channel `fp8_dynamic_quant` and shows ~20% throughput improvement; the change set is small and self-contained (single op + bench update). 
Anyone from the review group able to take a look so I can move on with downstream PRs that depend on this kernel path?
UPD: waiting for full bench resualts

PR #1539, opened 2026-06-10
Cherry-pick of Reapply path for block_size for non-GDN hybrid models to 0.21.0.post1

rsmyrek:
Changes:

Restored self.block_size = cache_config.block_size in the num_mamba_like_layers > 0 path of initialize_kv_cache. This was removed by PR https://github.com/vllm-project/vllm-gaudi/pull/1434 but is needed for non-GDN hybrids (e.g. GraniteMoeHybrid).

Added bucketing_manager.block_size = attn_block_size after bucketing initialization. This ensures decode buckets use the kernel granularity regardless of the KV-manager page size — fixing Qwen3.5 while keeping GraniteMoeHybrid correct.


jbyczkow (approved):
LGTM

PR #1552, opened 2026-06-17
Fix decode bucket sparsity for longprompt

shepark:
Add extra exponential buckets when max_decode_blocks exceeds max_blocks*3 to maintain bucket density at high block counts, avoiding excessive
padding that degrades throughput for large context workloads.

In case qwen3.5 122b-a10b model, there're perf regression with longprompt (ex: 200k) due to [f24f3f9d ](https://github.com/vllm-project/vllm-gaudi/commit/f24f3f9d) which causing overly wide max_decode_blocks, resulting excessive padding, but gives good low-end density and perf.

This is the test result with 122b-a10b model.

|Workload | v0.19.1 | v0.21.0 (before) | v0.21.0 (with fix) | Fix vs v0.19.1|
-- | -- | -- | -- | --
51200/2560 | 250.68 | 255.16 | 253.53 | +1.1%
102400/5120 | 212.36 | 223.25 | 218.57 | +2.9%
204800/10240 | 129.87 | 125.15 | 134.06 | +3.2%

With 35b-a3b model, 200k

|Workload | v0.19.1 | v0.21.0 (before) | v0.21.0 (with fix) | Fix vs v0.19.1| Fix vs v0.21.0|
-- | -- | -- | -- | -- | --
204800/10240 | 124.55 | 123.19 | 160.81 | +29.1% | +30.5%

And, corresponding changes in bucket test.

### DIFF
--- a/vllm_gaudi/extension/bucketing/exponential.py
+++ b/vllm_gaudi/extension/bucketing/exponential.py
@@ -92,6 +92,11 @@ class ExponentialBucketingStrategy():
         max_decode_blocks = math.ceil(max_model_len / block_size) * max_num_seqs
         max_decode_blocks = min(max_blocks, max_decode_blocks) if use_contiguous_pa else max_decode_blocks
         decode_blocks_limit = math.ceil(math.log2(max_decode_blocks)) + 1
+        # Compensate for wider range: add extra buckets to maintain density at high end
+        if not use_contiguous_pa and max_blocks > 0:
+            bounded_max = max_blocks * 3
+            if max_decode_blocks > bounded_max:
+                decode_blocks_limit += math.ceil(math.log2(max_decode_blocks / bounded_max)) + 1
         decode_block_bucket_cfg = [1, max_num_seqs, max_decode_blocks, decode_blocks_limit]
 
         msg = ("Decode bucket config (min, step, max_warmup, limit) "
--- a/tests/unit_tests/test_bucketing.py
+++ b/tests/unit_tests/test_bucketing.py
@@ -352,8 +352,9 @@ _REAL_BUGGY_MAX_DECODE_BLOCKS = 183808  # min(91964//128*256, 3593*256//4)
 @patch('vllm_gaudi.extension.bucketing.exponential.get_config')
 def test_real_scenario_decode_cfg_matches_fixed_log(mock_get_config):
     """Verify decode bucket config matches expected values for real scenario.
-    With non-contiguous PA: block config should be
-    [1, 256, ceil(91964/128)*256, ceil(log2(that))+1]
+    With non-contiguous PA: block config limit includes extra buckets
+    when max_decode_blocks > max_blocks*3 to maintain bucket density
+    at high block counts (longprompt density fix).
     """
     mock_get_config.return_value = _MockConfig(use_contiguous_pa=False)
     strategy = ExponentialBucketingStrategy()
@@ -366,6 +367,11 @@ def test_real_scenario_decode_cfg_matches_fixed_log(mock_get_config):
 
     expected_max = math.ceil(_REAL_MAX_MODEL_LEN / _REAL_BLOCK_SIZE) * _REAL_MAX_NUM_SEQS
     expected_limit = math.ceil(math.log2(expected_max)) + 1
+    # Account for extra buckets added when max_decode_blocks exceeds
+    # bounded_max (longprompt density compensation)
+    bounded_max = _REAL_MAX_BLOCKS * 3
+    if expected_max > bounded_max:
+        expected_limit += math.ceil(math.log2(expected_max / bounded_max)) + 1
     assert block_cfg[0] == 1, f"block min: expected 1, got {block_cfg[0]}"
     assert block_cfg[1] == _REAL_MAX_NUM_SEQS, (f"block step: expected {_REAL_MAX_NUM_SEQS}, got {block_cfg[1]}")
     assert block_cfg[2] == expected_max, (f"block max: expected {expected_max}, got {block_cfg[2]}")


vllm_gaudi/extension/bucketing/exponential.py
```diff
@@ -92,6 +92,11 @@ def get_decode_cfgs(self, max_num_seqs, block_size, max_num_batched_tokens, max_
         max_decode_blocks = math.ceil(max_model_len / block_size) * max_num_seqs
         max_decode_blocks = min(max_blocks, max_decode_blocks) if use_contiguous_pa else max_decode_blocks
         decode_blocks_limit = math.ceil(math.log2(max_decode_blocks)) + 1
+        # Compensate for wider range: add extra buckets to maintain density at high end
+        if not use_contiguous_pa and max_blocks > 0:
+            bounded_max = max_blocks * 3
```
yangulei, line 97: This hard-coded upper bound will truncate the decode block buckets which will cause re-compilation for some cases. For an example with `max_blocks=2000`, max_model_len=16384`, `max_num_seqs=128` and `block_size=128`. The decoding case with `context_lens = [8065, 1, 1, 1, ... 1, 1, 1]` needs `math.ceil(max(context_lens) / block_size) * max_num_seqs = ceil(8065 / 128) * 128= 8192` blocks after padding, which exceeds the upper bound of `max_blocks * 3 = 6000` and cause re-compilation for every following decoding steps for this batch.

yangulei (commented):
The performance drop for long `max_model_len` is caused by the fact that the relative padding (padded_value/bucket_value) towards to 50% for the exponential bucketing strategy. That's why I introduced the padding-aware bucketing strategy in https://github.com/vllm-project/vllm-gaudi/pull/762.

vllm_gaudi/extension/bucketing/exponential.py
shepark: @yangulei thank you for the review.
First off, this "bounded_max" is not the cap, but just a trigger that watching the bucket density.
The cap is still `max_decode_blocks = math.ceil(max_model_len / block_size) * max_num_seqs` which was defined by f24f3f9d.
With your config,
```
decode_bs_limit = ceil(log2(128)) + 1 = 7 + 1 = 8
decode_bs_bucket_cfg = [1, 2, 128, 8]
decode_query_bucket_cfg = [1, 1, 1, 1]
max_decode_blocks = ceil(16384 / 128) * 128 = 128 * 128 = 16,384
# use_contiguous_pa = False -> min() : not applied
max_decode_blocks = 16,384  (same as before)
decode_blocks_limit = ceil(log2(16384)) + 1 = 14 + 1 = 15
# with my PR:
bounded_max = 2000 * 3 = 6,000
max_decode_blocks(16,384) > bounded_max(6,000)? -> YES
decode_blocks_limit += ceil(log2(16384 / 6000)) + 1 = 3
decode_blocks_limit = 15 + 3 = 18
decode_block_bucket_cfg = [1, 128, 16384, 18]
```
variables | value | meaning
-- | -- | --
max_decode_blocks | 16,384 | bucket range cap
bounded_max | 6,000 | extra bucket trigger threshold
decode_blocks_limit | 18 (15+3) | number of exponential bucket
decode_block_bucket_cfg | [1, 128, 16384, 18] | final config

So, blocks = 8192 -> 8192 < 16,384 (cap) .. since it's still lower than the cap, "no re-comilation"

shepark:
> The performance drop for long `max_model_len` is caused by the fact that the relative padding (padded_value/bucket_value) towards to 50% for the exponential bucketing strategy. That's why I introduced the padding-aware bucketing strategy in #762.

I agree that your padding-aware bucketing strategy in #762 addresses the inherent ~50% worst-case padding issue of exponential bucketing. However, I'm not sure how that relates to this PR.
This PR is "Not" attempting to solve the fundamental limitation of exponential bucketing. It's fixing a throughput regression introduced by f24f3f9d, where max_decode_blocks was widened from max_blocks * 3 to ceil(max_model_len / block_size) * max_num_seqs without increasing the bucket count — resulting in sparser high-end spacing and excessive padding for `long-context workloads (200K+)` and resulting performance drop. The fix simply restores bucket density to pre-regression levels by adding extra buckets proportional to the overshoot.

vllm_gaudi/extension/bucketing/exponential.py
yangulei: Thanks for the explanation.
I see, the `max_decode_blocks` keep unchanged. This PR just add several buckets if necessary.

yangulei:
> > The performance drop for long `max_model_len` is caused by the fact that the relative padding (padded_value/bucket_value) towards to 50% for the exponential bucketing strategy. That's why I introduced the padding-aware bucketing strategy in #762.
> 
> I agree that your padding-aware bucketing strategy in #762 addresses the inherent ~50% worst-case padding issue of exponential bucketing. However, I'm not sure how that relates to this PR. This PR is "Not" attempting to solve the fundamental limitation of exponential bucketing. It's fixing a throughput regression introduced by [f24f3f9](https://github.com/vllm-project/vllm-gaudi/commit/f24f3f9df8c10c1d03b489cba8a9294b75da95e3), where max_decode_blocks was widened from max_blocks * 3 to ceil(max_model_len / block_size) * max_num_seqs without increasing the bucket count — resulting in sparser high-end spacing and excessive padding for `long-context workloads (200K+)` and resulting performance drop. The fix simply restores bucket density to pre-regression levels by adding extra buckets proportional to the overshoot.

My bad, I thought this PR re-apply the WA of `max_decode_blocks = max_blocks * 3`.

adobrzyn (approved):

PR #1567, opened 2026-06-25
MXFP4 GPT-OSS SwiGLU-OAI + bias serving

osavchenkox:
## Summary
Route the GPT-OSS MXFP4 MoE through the native
`torch.ops.hpu.mixture_of_experts.bias_mxfp4_fused_weights` op so SwiGLU-OAI
(alpha=1.702, limit=7.0) + per-expert bias are applied on-device on the packed
weights, instead of dequantizing the experts to bf16 at load time.

## Changes
- `extension/ops.py`: `MoeMXFP4Matmul.set_bias`; `VllmMixtureOfExpertsOpMXFP4`
  gains `has_bias`/`alpha`/`limit`, caches w13/w2 bias views, and adds a
  bias-aware compiled forward calling `bias_mxfp4_fused_weights` (falls back to
  `mxfp4_fused_weights` when there is no bias).
- `ops/hpu_mxfp4.py`: detect `w13_bias`/`w2_bias` and load them into the op.
- `models/gptoss_mxfp4.py`: enable the `HPU_MXFP4_NATIVE` branches (were hardcoded
  off) so packed uint8 weights + E8M0 scales + bias reach the native op; and fix
  the reference (`HPU_MXFP4_NATIVE=0`) loader to call the branch's own
  `_load_weights_mxfp4_dequantize_hpu` (the old `_load_weights_mxfp4` name fell
  through to the upstream FusedMoE loader and raised
  `shard_id must be ['w1','w2','w3'] but got None`).

## Validation
- Op A/B on Gaudi3: cosine 0.997 vs bf16.
- End-to-end gpt-oss-20B serve A/B (1× Gaudi3, vLLM, real packed-mxfp4 checkpoint):
  native (`HPU_MXFP4_NATIVE=1`) vs bf16 dequant-at-load (`=0`) on the **same**
  checkpoint = **bit-identical** output (10/10 raw prompts + chat "...Paris.",
  1142 tokens). Native arm confirmed `quantization=gpt_oss_mxfp4` + `hpu_backend`.
- Follow-up (not a blocker): 120B multi-card and a published lm_eval score are not
  yet done.

## Test plan
- [x] Op A/B on Gaudi3: cosine 0.997.
- [x] gpt-oss-20B end-to-end serve A/B: bit-identical.
- [ ] 120B multi-card (follow-up).

## Dependencies
- A companion `habana_frameworks` build providing the
  `bias_mxfp4_fused_weights` op, plus the complex_guid MoeExperts GPT-SwiGLU
  MXFP4 fix. All three land together.
- Built on top of **#1526 "Add Single Card support in GPTOSS"** (open, not
  merged). This PR currently targets `main`, so its diff includes #1526's two
  commits until #1526 lands; GitHub drops them automatically once #1526 merges.
  **Kept as a draft until #1526 is merged** — then rebase onto `main` and mark
  ready.


vllm_gaudi/extension/ops.py
```diff
@@ -1522,6 +1522,204 @@ def forward(
...
+                        d_scale_w3,
+                        block_size=block_size,
+                        permuted_weights=True,
+                        activation="silu",
+                        experts_min=experts_min,
+                        experts_max=experts_max,
+                        is_fp4=True,
+                        chunk_size=chunk_size,
+                        total_experts=total_experts,
+                    )
```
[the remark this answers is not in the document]
osavchenkox: Fixed in 623d1dc0 — the non-bias `_mxfp4_fused_fwd` now passes `activation` through to `mxfp4_fused_weights` instead of hardcoding `"silu"`. `forward()` already normalizes the value via `_as_activation_str` before the call, so the bias and non-bias paths now behave the same.

vllm_gaudi/ops/hpu_fused_moe.py
```diff
@@ -140,7 +140,7 @@ def apply_monolithic(
 
         topk_ids = topk_ids.view(-1, topk_ids.shape[-1])
         topk_weights = topk_weights.view(-1, topk_weights.shape[-1])
-
+        # print(f"Layer Name: {layer}, Op name: {layer.moe_op} and activation name: {layer.activation}")
```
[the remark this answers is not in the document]
osavchenkox: Agreed in principle. This line is part of the `WIP: Temporary Workaround to use mxfp4 fused moe kernel` commit from #1526, which this branch is stacked on — it is not introduced by this PR and will drop out of the diff once #1526 merges. Leaving it to be addressed in #1526 to avoid editing its commits here.

vllm_gaudi/ops/hpu_mxfp4.py
```diff
@@ -0,0 +1,157 @@
...
+        super(GptOssMxfp4MoEMethod, self).__init__(moe)
+        self.weight_dtype = "gpt_oss_mxfp4"
+        # No CUDA backend needed
+        self.mxfp4_backend = None
+        self.experts_cls = None
+        self.moe_kernel = None
+        self._cache_permute_indices = {}
+
+        self.use_dispatch_fn = get_config().use_dispatch_fn
+        torch.hpu.synchronize()
```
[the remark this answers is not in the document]
osavchenkox: Agreed. This `torch.hpu.synchronize()` comes from the `Add Single Card support in GPTOSS` commit in #1526, which this branch is stacked on — not introduced by this PR. It will leave this diff once #1526 merges, so best addressed there rather than editing #1526's commit from here.

SKRohit:
LGTM can be merged.

vllm_gaudi/ops/hpu_fused_moe.py
```diff
@@ -204,7 +204,7 @@ def apply_monolithic(
...
```
SKRohit: please remove this comment

SKRohit: Also add changes present in this commit: https://github.com/vllm-project/vllm-gaudi/pull/1578/changes/ef9a4d367d5f110fc228cb12423513755a289bf8

osavchenkox: Done — removed the commented-out debug print in 501503aa.

osavchenkox: Ported the changes from #1578's commit ef9a4d3 ("fix gptoss bf16 path") in 501503aa: `_load_weights_mxfp4_dequantize_hpu` now iterates via `remap_moe_expert_weights(weights, params_dict)`, and both the w13 and w2 block-weight lookups fall back to the `.experts.routed_experts.` name before raising. I applied the w2 branch without the duplicated nested `if` that's in the original commit — same behavior, cleaner.

adobrzyn (approved):

iboiko-habana:
Fix MXFP4 plugin import after upstream FusedMoE rename

Issue Description
The MXFP4 implementation ported by vllm-gaudi PR #1567 imported FusedMoE from:
This worked with vLLM v0.26.0, where FusedMoE remained available as a factory function. Upstream vLLM PR #44941 renamed that function to FusedMoEFactory.
On current vLLM main, loading the HPU plugin therefore failed during model architecture inspection. After correcting the import, the apply_monolithic annotation still referenced FusedMoE, causing:
Because plugin loading occurs in the model registry subprocess, this appeared as unrelated ModelConfig validation failures for models such as OPTForCausalLM.

Fix
diff --git a/vllm_gaudi/ops/hpu_mxfp4.py b/vllm_gaudi/ops/hpu_mxfp4.py
--- a/vllm_gaudi/ops/hpu_mxfp4.py
+++ b/vllm_gaudi/ops/hpu_mxfp4.py
@@ -103,7 +103,7 @@ class HPUGptOssMxfp4MoEMethod(GptOssMxfp4MoEMethod):
 
     def apply_monolithic(
         self,
-        layer: FusedMoE,
+        layer: RoutedExperts,
         x: torch.Tensor,
         router_logits: torch.Tensor,
         **kwargs,

PR #1574, opened 2026-06-28
[kimi2.6] enable model text only initial

sureshnam:
Adds VLLM_SKIP_MM_WARMUP=1 to skip only the multimodal-graph warm-up while still running the prompt and decode graph warm-up.

Motivation: some multimodal models include vision towers whose upstream implementation hits a torch._inductor code path that has no 'hpu' backend registered, causing engine init to crash in warmup_multimodal_graphs. Today the only workaround is VLLM_SKIP_WARMUP, which disables all warmup and causes runtime JIT-compile stalls on every new bucket size hit. With this knob, users running such models text-only can keep the normal prompt/decode bucket warmup and avoid the runtime stalls, while skipping just the vision compile that they don't need.

Pattern follows existing in-file env-var gates such as VLLM_SKIP_MARK_UNBACKED. Documents the new variable under Developer Mode Parameters next to VLLM_SKIP_WARMUP.

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -5669,8 +5669,17 @@ class HPUModelRunner(HpuKVConnectorModelRunnerMixin):
         # to avoid GC errors. In torch.compile mode, run it inside
         # for faster recipe-only compilation.
         use_torch_compile = (not htorch.utils.internal.is_lazy() and not self.model_config.enforce_eager)
-        if self.supports_mm_inputs and not use_torch_compile:
+        # VLLM_SKIP_MM_WARMUP: when set, skip ONLY the multimodal-graph warmup
+        # while still running prompt+decode warmup. Useful for models whose
+        # vision tower fails to compile on HPU (e.g. when an upstream module
+        # uses a torch._inductor path without an 'hpu' backend), and when the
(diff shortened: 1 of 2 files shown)


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -5669,8 +5669,17 @@ def warmup_model(self) -> None:
...
         # for faster recipe-only compilation.
         use_torch_compile = (not htorch.utils.internal.is_lazy() and not self.model_config.enforce_eager)
-        if self.supports_mm_inputs and not use_torch_compile:
+        # VLLM_SKIP_MM_WARMUP: when set, skip ONLY the multimodal-graph warmup
+        # while still running prompt+decode warmup. Useful for models whose
+        # vision tower fails to compile on HPU (e.g. when an upstream module
+        # uses a torch._inductor path without an 'hpu' backend), and when the
+        # user is running the model text-only and does not need the vision
+        # path. Differs from VLLM_SKIP_WARMUP, which disables all warmup.
+        skip_mm_warmup = os.environ.get('VLLM_SKIP_MM_WARMUP', '0').lower() in ('1', 'true')
```
adobrzyn, line 5678: Can we add it to features and use get_config?

iboiko-habana:
@sureshnam  do we need this PR or it can be closed

PR #1579, opened 2026-06-30
granite-4.0-h-small - remove OOM workaround

rsmyrek:
This pull request removes the logic that capped the prompt warmup context size for non-GDN Mamba-hybrid models in the `_prepare_dummy_scenario` method. Previously, this cap was used to prevent out-of-memory (OOM) issues during model warmup by limiting the number of context blocks, but this logic and its warning message have now been deleted.

**Prompt warmup context handling:**

* Removed the block of code that limited the number of prompt context blocks (`prompt_num_blocks`) for non-GDN Mamba-hybrid models, including the associated warning message and explanatory comments. This means prompt warmup will no longer be artificially capped for these models.


jbyczkow (approved):

jbyczkow (approved):

wpyszka (approved):
approved for 0.24.0

PR #1588, opened 2026-07-02
Support HPU-prefill to GPU-decode NIXL KV transfer (joint-KV staging)

hsubramony:
In vLLM v0.24.0 the NIXL connector registers KV cache as one joint [K|V] region per layer for blocks-first backends (e.g. GPU FLASH_ATTN with HND layout), reading V at block_len // 2 within each block. The HPU hetero connector still advertised separate K and V regions, so a GPU decode instance rejected the prefill handshake with "Number of KV layers must match between prefill and decode".

Add an opt-in joint-KV staging path (VLLM_HPU_NIXL_JOINT_KV) that lets an HPU prefill instance serve a GPU decode instance:

- Register a bounded staging buffer laid out blocks-first [num_slots, 2(K,V), n_kv_heads, block_size_on_save, head_size] and advertise one joint region per layer, matching the consumer's region model. On save, transform each request's KV blocks (NHD->HND, logical->on-save sub-blocks) and copy K/V into assigned staging slots.
- Allocate staging slots on the scheduler, carry them to the worker via connector metadata, advertise them as remote_block_ids, and free them on transfer completion (decode fetch notification).
- Size the pool deterministically from config (max_num_seqs * ceil(max_model_len / block_size_on_save), VLLM_HPU_NIXL_STAGING_SLOTS overrides) and fail fast at registration if it does not fit device memory, with actionable guidance.
- Block counts use num_prompt_tokens (stable across chunked-prefill steps and equal to the decode's num_external_tokens) so the scheduler, worker, and decode agree; a mismatch previously tripped the decode's num_local_blocks <= remote assertion.
- Under load, requests hold slots until the decode fetches them, so the pool can be momentarily full: defer such saves (KV kept on the HPU, retried on a later step) instead of failing. If a request still lacks a reservation at finish, fail only that request rather than crashing the engine.

Store kv_cache_config on the worker (the base connector sets it in its own register_kv_caches, which this override replaces).

Verified end to end: an HPU prefill instance streams KV to a GPU decode instance and produces correct completions under concurrent load.


libinta:
when Gaudi is the decode (pull) worker, the NIXL handshake crashes:
AttributeError: 'NixlPullConnectorWorker' object has no attribute '_engine_ttl'
→ EngineDeadError → 500 Internal Server Error

Why this branch doesn't surface it: the target scenario here is HPU-prefill → GPU-decode, where the decode/pull worker is a GPU running stock vLLM (which sets _engine_ttl normally). The crash only appears when Gaudi is the decode — i.e. the HPU↔HPU hetero path.

Root cause: vLLM PR #35264 (in v0.24.0) split the NIXL connector into base_worker + pull/push workers and added _engine_ttl/_evict_stale_engines, region-MLA tracking, HMA, etc. The gaudi connector monkey-patches only __init__/register_kv_caches/_read_blocks, so the non-overridden upstream methods (_ensure_handshake → _evict_stale_engines, add_remote_agent → _validate_remote_agent_handshake) read attributes the gaudi __init__ override never sets: _engine_ttl, _engine_last_active, attn_backends, _layer_specs, hma_group_size, _conv_decomp, _region_is_mla, _logical_num_blocks, num_descs. (This PR already adds kv_cache_config.)

Repro (single node, both HPU, pinned vLLM):
# prefill TP1 bs128, decode TP1 bs256, NHD both sides, enable_permute_local_kv
VLLM_HPU_HETERO_KV_LAYOUT=true ... kv_buffer_device=hpu, enable_permute_local_kv=true
# decode's first start_load_kv() → _ensure_handshake → _evict_stale_engines → self._engine_ttl
Fix I validated — add the 9 attributes (init defaults + register-time values), plus make remote_block_ids nesting block_size_ratio-dependent on the non-joint path (flat when ratio==1, wrapped when >1), leaving your joint-KV staging wrap untouched.

Results on Gaudi (Llama-3.1-8B, gsm8k limit=128): with these fixes, hpu-GDR homogeneous + hetero receiver (bs128→bs256) + hetero sender (bs256→bs128) all pass. Without them, the two hetero cases crash on _engine_ttl.

I have libinta/vllm-gaudi:libint/hetero_combined to combine the fix + your PR 1588, please check. you can update your PR with it. Also add the test cases I put in your PR

PR #1594, opened 2026-07-03
Port: Fix Maverick FP8 INC calibration OOM after #41184 MoE refacotr #1590

iboiko-habana:


adobrzyn:
Heads-up on porting order: #1605 (GAUDISW-250139, dense-model INC OOM fix) gates the MoE pre-stage `.to('hpu')` + `_rebind_moe_expert_weights` that this PR introduces to main. It should be ported **right after this lands** — cherry-picking it before #1594 conflicts (main's INC block is still the pre-#1590 single unconditional `.to('hpu')`), and the regression it fixes doesn't exist on main until this merges. Tracking the #1605 port on my side.

adobrzyn (approved):

PR #1596, opened 2026-07-05
[ATEA] [google/gemma-4-31B] Gaudi3 /hpu support

sureshnam:
TP=2 MAX_LEN=131072 MAX_SEQS=32  ./run_gemma4.sh   # two cards, 128K context

Effective command:

```bash
PT_HPU_ENABLE_LAZY_COLLECTIVES=true \
vllm serve google/gemma-4-31B --tensor-parallel-size $TP --max-model-len $MAX_LEN \
    --dtype bfloat16 --max-num-seqs $MAX_SEQS --gpu-memory-utilization 0.9 --port 8000
```

Notes:
- TP=1 fits on one 128 GB card but leaves only ~65K KV-cache tokens
  (≈32 concurrent 2K-token requests); TP=2 shards weights and gives
  185K KV tokens.
- **128K context requires `MAX_SEQS=32`** — with the default 128 the largest
  decode warmup bucket OOMs during HPU graph compilation.
- Warmup takes ~7 min at TP=1/8K, ~11 min at TP=2/128K.
- lm-eval in this container needs `pip install tenacity aiohttp`.

## Accuracy

`lm-eval 0.4.12`, full GSM8K test set (1,319 questions), 5-shot,
greedy (`temperature=0`), via `local-completions` against the live TP=1 server:

| Task | n-shot | Filter | Metric | Value |
|---|---|---|---|---|
| gsm8k | 5 | strict-match | exact_match | **0.8393 ± 0.0101** |
| gsm8k | 5 | flexible-extract | exact_match | **0.8393 ± 0.0101** |

Strict and flexible extraction agree, and the score is in the expected range for
a strong 31B base model — confirming the head-dim-512 attention path is
numerically correct (broken attention would collapse this to ~0).

### TP=2, max_model_len 131072, 4K ISL / 1K OSL
Config: `TP=2 MAX_LEN=131072 MAX_SEQS=32`. KV cache: **185,728 tokens**
(2.9× TP=1) — fits a full 128K-context request or ~36 concurrent 4K+1K requests.
`max_num_seqs` must be lowered to 32 at this context length: the default 128
makes the largest decode warmup bucket (batch 128 × 128K-token block range)
exceed device memory (`PT_DEVMEM OOM: No enough memory for defragment` during
graph warmup). Warmup takes ~11 min (645 s, 16.6 GiB graph memory).

| Metric | Concurrency 32 (96 reqs) |
|---|---|
| Total token throughput | **2,885 tok/s** (2,308 in + 577 out) |
| Output token throughput | 577 tok/s |
| Peak output throughput | 768 tok/s |
| Request throughput | 0.56 req/s |
| Median TTFT | 1.13 s |
| Mean / P99 TTFT | 4.59 s / 18.2 s (wave starts) |
| Median TPOT | 52.0 ms (~19 tok/s per stream) |
| P99 TPOT | 56.1 ms |
| Prefill throughput (observed peak) | ~8,600 tok/s |
| Failed requests | 0 / 96 |

No preemptions at steady state (KV usage ≈ 79%). TPOT is flat (P99 within 8% of
median) — decode is compute-bound and stable at this batch size.


iboiko-habana:
@sureshnam do we need this PR or it can be closed?

sureshnam:
closed as other PR complete the work

PR #1601, opened 2026-07-06
[ATEA] [IBM] Gemma4 e4b model enable

sureshnam:
vllm_gaudi/v1/worker/hpu_model_runner.py`, `get_attn_backends_for_group`.
KV-sharing layers are appended to their target layer's KV cache group
(`add_kv_sharing_layers_to_kv_cache_groups`) but own no entry in the merged
`UniformTypeKVCacheSpecs.kv_cache_specs` dict, so the per-layer spec unwrap
crashed at engine init with `KeyError: '...layers.24.self_attn.attn'` (layer 24
= first shared layer). GPU never hits this because its hybrid KV cache manager
keeps sliding/full layers in separate groups (no uniform merge). Fix: resolve
missing layers through `self.shared_kv_cache_layers[layer_name]` to the target
layer's spec (identical KV layout by definition).

```bash
PT_HPU_ENABLE_LAZY_COLLECTIVES=true \
vllm serve google/gemma-4-E4B --tensor-parallel-size 1 --max-model-len 32768 \
    --dtype bfloat16 --max-num-seqs 32 --gpu-memory-utilization 0.9 --port 8000
```

KV cache: **1,689,472 tokens** at these settings �~@~T KV sharing, 2 KV heads, and
the 512 sliding window make this model exceptionally KV-cheap; the full 131K
context at TP=1 is trivially feasible. Warmup �~I~H 21 min (165 buckets).

## Accuracy

lm-eval 0.4.12, **full GSM8K test set** (1,319 questions), 5-shot, greedy, via
the live server �~@~T the whole eval completed in ~100 seconds:

| Task | Filter | exact_match |
|---|---|---|
| gsm8k (5-shot) | strict-match | **0.4496 ± 0.0137** |
| gsm8k (5-shot) | flexible-extract | **0.4534 ± 0.0137** |

In the expected band for a ~4B-effective base model; smoke test answers
"6�~W7 �~F~R The answer is 42." correctly.

## Performance

`vllm bench serve`, random dataset, 4K ISL / 1K OSL, `--ignore-eos`:

| Metric | Concurrency 32 (96 reqs) |
|---|---|
| Total token throughput | **4,622 tok/s** (3,698 in + 924 out) |
| Output token throughput | 924 tok/s (peak 1,824) |
| Request throughput | 0.90 req/s |
| Median TTFT | 362 ms |
| Median TPOT | 20.96 ms (~48 tok/s per stream) |
| P99 TPOT | 31.76 ms |
| Failed requests | 0 / 96 |


sureshnam:
Closed PR as official work PR in progress

PR #1602, opened 2026-07-07
Qwen3.5 MOE: Fix post-warmup graph recompilation

jiminha:
Dynamo's newest-first guard evaluation combined with symbolic shape guards caused post-warmup recompilations. Newer compilations for smaller seq_lens would shadow older compilations that had recipes for max seq_len shapes.

Fixes:
- Add validation warmup for max seq_len buckets at end of warmup
- Add qwen3_5_moe to flatten_input feature (3D→2D tensor reshape)
- Add HpuQwen3NextModel with residual=zeros_like instead of None
- Add _init_moe_quant_config() for eager MOE quant initialization

The validation warmup ensures the newest dynamo compilations have Synapse recipes for the largest bucket shapes that are processed first during initial warmup but get shadowed by later symbolic-guard compilations.

 | Config | Warmup | Memory | Run1 tok/s | Run2 tok/s |
|--------|--------|--------|------------|------------|------------|
| **No fixes (original)** | 948s (~16m) | 4.86 GiB | **222** | 256 | 
| **With fix** | **988s (~16m)** | **4.39 GiB** | **251** | 253 |


jiminha:
@adobrzyn  Please review this code. 
For Qwen3.5, all buckets for the max seq_len validation warmup are needed. The top 2 (largest num_blocks) trigger dynamo recompilation and establish the newest compilation (graph_0048). The remaining buckets (3–11) still need to run since they create Synapse recipes in that newest dynamo compilation — which is the one inference will actually select. Without them, first inference for each num_blocks value pays a multi-second Synapse compile penalty.

adobrzyn (approved):

PR #1607, opened 2026-07-08
Warn when calibrate_model.sh -u is used on TP>1 without -r

adobrzyn:
## Problem
With `-u`, the measure/quantize steps (2 and 4) run with expert parallelism, so on `TP>1` each rank only measures its **EP-local** experts (e.g. 16 of 128 for Llama-4 Maverick on TP8). The unify step (step 5) merges these per-rank measurements back into the full expert set, but it runs **only when a target rank is given via `-r`**.

Passing `-u` on `TP>1` without `-r` therefore leaves the measurements covering only each rank's local experts. For models with many experts split across ranks (Llama-4 Maverick, DeepSeek), the non-local experts get no dedicated scale and fall back to coarse quantization at serve time, which lowers FP8 accuracy. There is no error today — just a silently worse result. (Small-expert models such as Mixtral are unaffected.)

## Change
- Emit a warning when `-u` is set with `TP>1` but `-r` is missing.
- Clarify the `-u` help text (it also enables expert parallelism in steps 2/4; unification needs `-r`).
- Document the unify requirement for MoE models in the calibration docs.

No behavior change to existing valid workflows.

### DIFF
--- a/calibration/calibrate_model.sh
+++ b/calibration/calibrate_model.sh
@@ -22,7 +22,7 @@ usage() {
     echo "  -l    - limit number of samples in calibration dataset"
     echo "  -t    - tensor parallel size to run at (default: 1); NOTE: if t > 8 then we need a multi-node setup"
     echo "  -r    - rank of unified measurements, it should be smaller than original rank number and should be a factor of the original rank number"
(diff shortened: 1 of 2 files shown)


mfylcek (approved):
LGTM

mgawarkiewicz-intel (approved):

PR #1624, opened 2026-07-18
Fix GraniteMoeHybrid GDN misdetection from transformers>=5 layer-type remap

rsmyrek:
This pull request addresses an issue with model type detection for certain hybrid models in the `hpu_model_runner.py` initialization logic. Specifically, it prevents misclassification of non-GDN Mamba hybrid models (like Granite 4.0-H) as GDN/linear-attention models, which previously led to incorrect cache initialization and non-deterministic behavior.

**Model type detection improvements:**

* Added a `_non_gdn_mamba_hybrid_model_types` tuple to explicitly list model types (e.g., `"granitemoehybrid"`) that should not be treated as GDN/linear-attention models, even if their Mamba layers are relabeled by newer versions of transformers.
* Updated the logic for setting `self.num_gdn` to skip these non-GDN hybrid types, ensuring correct cache alignment and deterministic behavior for affected models.… remap

transformers>=5's remap_legacy_layer_types renames GraniteMoeHybrid's plain mamba layers from "mamba"/"attention" to "linear_attention"/ "full_attention" on config load. get_num_layers_by_block_type(..., "linear_attention") then matches these relabeled layers, so num_gdn is misdetected as >0, which skips the block_size=528 alignment for Granite 4.0-H and re-triggers "not warmed-up" recompilations and non-deterministic tool-calling output.

Exclude granitemoehybrid from the GDN auto-detection by model_type so its plain mamba layers are no longer counted as GDN/linear-attention.


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -1258,8 +1258,21 @@ def __init__(
...
+        # relabeled mamba layers, so num_gdn is misdetected as >0 below. That
+        # skips the block_size=528 alignment for Granite 4.0-H (see
+        # initialize_kv_cache) and re-triggers "not warmed-up" recompilations /
+        # non-deterministic tool-calling output.
+        _non_gdn_mamba_hybrid_model_types = ("granitemoehybrid", )
+
         self.num_gdn = 0
-        if self.num_mamba_like_layers > 0:
+        if (self.num_mamba_like_layers > 0
+                and self.model_config.hf_config.model_type not in _non_gdn_mamba_hybrid_model_types):
```
jbyczkow: Condition be replaced with:
and self.model_config.hf_config.model_type != "granitemoehybrid" 
and _non_gdn_mamba_hybrid_model_types can be removed

iboiko-habana (approved):

PR #1632, opened 2026-07-23
Gemma4:optimize sliding window

libinta:
1. Problem: Gemma-4 has 50 sliding-window attention layers and only 10 full-attention layers. Before this change, prefill attention for every layer — sliding-window or not — gathered the entire context block list, even though sliding-window layers only ever attend to the last window_size tokens. That wastes both compute (FusedSDPA processes irrelevant KV blocks) and memory (the attention mask materializes the full context width).

Fix — window_block_list: For sliding-window layers, build a second, narrower block list containing only the last window_size // block_size context blocks per request, and route those layers through it instead of the full block_list.

test gemma-4-E4B, prompt=12,016 tokens, max_num_batched_tokens=2048, sliding_window=512). The output text was correct/coherent
Memory optimization: the optimization fired 5 times during prefill (mask sites) — here are the actual mask-memory numbers per chunk:
| Chunk | full_ctx_cols (untrimmed) | trimmed max_context_len | attn_bias actual (TRIMMED) | Would-be size (UNTRIMMED) | Saved |
|-------|---------------------------|--------------------------|------------------------------|-----------------------------|--------------|
| 1     | 3,840                     | 512                      | 10.00 MB                     | 23.00 MB                    | 13 MB (57%)  |
| 2     | 7,424                     | 512                      | 10.00 MB                     | 37.00 MB                    | 27 MB (73%)  |
| 3     | 7,424                     | 512                      | 10.00 MB                     | 37.00 MB                    | 27 MB (73%)  |
| 4     | 14,336                    | 512                      | 10.00 MB                     | 64.00 MB                    | 54 MB (84%)  |
| 5     | 14,336                    | 512                      | 10.00 MB                     | 64.00 MB                    | 54 MB (84%)  |

Compute: block_list trimmed to a constant 4 blocks (window 512 / block 128) regardless of context length — logged real trims from 16→4 up to 125→4 (≈31× fewer KV blocks gathered on the hottest layers).

End-to-end throughput (real A/B vs parent commit, vllm bench serve, E4B-it, 6300 input + 20 images + 380 output):
~8% median TTFT 17%-8% improvement
2. Replace transformers sdpa_attention_forward with HPU-optimized version
that uses FusedSDPA kernel when fsdpa_impl is configured. This provides
1.4-2x performance improvement for models using transformers SDPA
attention (e.g., vision encoders in Gemma4, LLaVA, Qwen2-VL).

### DIFF
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -1654,6 +1654,16 @@ class HPUModelRunner(HpuKVConnectorModelRunnerMixin):
                 kv_cache_spec[layer_name] = attn_module.get_kv_cache_spec(self.vllm_config)
             elif isinstance(attn_module, Attention):
                 if attn_module.attn_type == AttentionType.DECODER:
+                    # Normalize page_size_bytes across heterogeneous head_size layers
+                    # (e.g. Gemma4: sliding head_size=256, full head_size=512)
+                    max_head_size = max(
+                        m.head_size for m in forward_ctx.values()
+                        if isinstance(m, Attention)
+                        and m.attn_type == AttentionType.DECODER
+                        and getattr(m, 'kv_sharing_target_layer_name', None) is None
+                    )
+                    if attn_module.head_size < max_head_size:
+                        layer_block_size = block_size * (max_head_size // attn_module.head_size)
                     kv_cache_spec[layer_name] = FullAttentionSpec(block_size=block_size,
                                                                   num_kv_heads=attn_module.num_kv_heads,
                                                                   head_size=attn_module.head_size,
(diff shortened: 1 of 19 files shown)


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -7192,15 +7207,28 @@ def _set_attn_bias_for_sliding_window(self, attn_metadata: HPUAttentionMetadataV
...
 
-            invalid_lens_t = context_lens_t - window_size + torch.arange(seq_len, device=device) - 1
-            past_indices = torch.arange(max_context_len, device=device)
-            past_mask = ((past_indices.unsqueeze(0) > invalid_lens_t.unsqueeze(-1)) &
-                         (past_indices.unsqueeze(0) < context_lens_t.unsqueeze(-1).unsqueeze(0))).unsqueeze(1)
+            if window_block_list is not None:
+                # Trimmed coordinate frame: gathered KV is the last `max_context_len`
+                # context tokens, so re-anchor context_lens_t to that frame before
+                # applying the same per-query-token window shift as the untrimmed path.
+                context_lens_trimmed = torch.clamp(context_lens_t, max=max_context_len)
```
adobrzyn: Quick Q:  The window trim keeps sliding_block_size = self.sliding_window // self.attn_block_size blocks (blocks[-sliding_block_size:]), and the mask re-anchors via context_lens_trimmed = torch.clamp(context_lens_t, max=max_context_len). Did you consider what happens when context_len isn't a multiple of attn_block_size?

Like context_len=600, block_size=128, sliding_window=512, keeps blocks[-4:] = tokens 128..599, but the clamp anchors as if index i -> token 88+i. That's off by 40 and drops in-window tokens 89..127 - wrong attention, but no crash

libinta: fixed based on your comment. thx

libinta:
dup of [16](https://github.com/vllm-project/vllm-gaudi/pull/1644)

PR #1634, opened 2026-07-24
[Do not merge] Enable Nemotron-H (TP8xPP2) on HPU

ldurejko:
Add pipeline-parallelism support to the HPU model runner and fix a set of Nemotron-H bring-up issues so the model runs as TP8xPP2 across 16 HPUs.

- hpu_model_runner: implement PP send/recv of IntermediateTensors between stages (with zeroed warmup tensors on non-first stages), and allocate separate contiguous tensors for mixed-dtype standard Mamba2 KV caches instead of aliased as_strided views (aot_autograd cannot compile in-place mutations on differently-typed views of one buffer during decode).
- hpu_mamba_mixer2: return the flat [num_tokens, hidden] output for flatten_input models (e.g. Nemotron-H) instead of the [batch, seq, hidden] reshape, so downstream MoE/MLP layers receive the 2D layout they expect.
- features: add nemotron_h to the flatten_input allowlist.
- hpu_fused_moe: support non-gated (is_act_and_mul=False) MoE activations.
- platform: auto-disable async scheduling under pipeline parallelism (the PP path consumes scheduler-provided new_token_ids, which async mode omits).


jkaniecki (changes requested):
This is not a full-fledged enablement. It's a big WA for real problem - It adds custom, unoptimized MoE up, which should be avoided - we should reuse out torch.hpu.ops.mixture_of_experts and modify it to match nemotrons calculations. However, we can base on this PR to primarily check acc and execution flow of the model. THIS SHOULDN'T BE MERGED

PR #1639, opened 2026-07-27
Add pure-Python MiniMax-M3 tool-call parser for HPU

mkrze:
## Summary

Adds an out-of-tree, pure-Python MiniMax-M3 tool-call parser (`minimax_m3_py`) for vLLM on HPU.

Upstream vLLM's built-in `minimax_m3` parser is implemented in Rust (the `vllm._rust_tool_parser` PyO3 extension), which is not compiled into the Gaudi `+empty` wheel and therefore 500s on HPU. This parser re-implements the same XML tool-call grammar in pure Python so it runs anywhere, registered under the distinct name `minimax_m3_py` to avoid colliding with the upstream Rust parser.

## Changes

- **New parser** (`vllm_gaudi/entrypoints/openai/tool_parsers/minimax_m3.py`): pure-Python implementation of the MiniMax-M3 XML tool-call grammar with truncation-safe incremental streaming (emits tool name first, streams argument deltas, and auto-closes truncated arguments on end-of-stream), fixing the "agent gets stuck after the discovery phase" hang. Any parse error degrades to returning content instead of a 500.
- **Plugin registration**: `register_tool_parsers()` in `vllm_gaudi/__init__.py`, wired via a new `04.hpu_tool_parsers` entry point in `setup.py`, so the parser is auto-registered at plugin load — no `--tool-parser-plugin` file required.
- **Docs**: quickstart section documenting how to enable MiniMax-M3 tool calling on HPU.
- **Tests**: unit tests in `tests/unit_tests/test_minimax_m3_tool_parser.py` covering parsing, fallbacks, and streaming behavior.

## Usage

```bash
export PT_HPU_ENABLE_LAZY_COLLECTIVES=true
export VLLM_ALLOW_LONG_MAX_MODEL_LEN=1
export VLLM_ENGINE_ITERATION_TIMEOUT_S=3600
export VLLM_RPC_TIMEOUT=100000
export VLLM_GRAPH_RESERVED_MEM=0.4        # shrink KV pool, free HBM for warmup graphs

export VLLM_MINIMAX_M3_MOE_DECODE_GATHER=1

vllm serve MiniMaxAI/MiniMax-M3 --tensor-parallel-size 8 --block-size 128 \
  --max-model-len 131072 --dtype bfloat16 --trust-remote-code \
  --gpu-memory-utilization 0.85 --reasoning-parser minimax_m3 --port 8000 \
  --max-num-seqs 256 --max-num-batched-tokens 8192 \
  --enable-auto-tool-choice \
  --tool-parser-plugin /software/users/mkrzemie/minimax-m3/minimax_m3_tool_parser_plugin_stream.py \
  --tool-call-parser minimax_m3_py \
  --default-chat-template-kwargs '{"thinking_mode": "disabled"}' \
  --override-generation-config '{"max_new_tokens": 8192}'

### DIFF
--- a/docs/getting_started/quickstart/quickstart_inference.md
+++ b/docs/getting_started/quickstart/quickstart_inference.md
@@ -105,6 +105,22 @@ Then, use the OpenAI Python client or curl:
         }'
     ```
 
+## MiniMax-M3 Tool Calling
+
+MiniMax-M3 uses an XML tool-call format. On HPU, select the pure-Python parser
+provided by `vllm-gaudi` because the upstream Rust parser is not included in the
+HPU `+empty` build:
+
+```bash
+vllm serve MiniMaxAI/MiniMax-M3 \
+    --reasoning-parser minimax_m3 \
+    --enable-auto-tool-choice \
+    --tool-call-parser minimax_m3_py
+```
+
+The `vllm-gaudi` plugin registers `minimax_m3_py` automatically; no
+`--tool-parser-plugin` argument is required.
+
 ## OpenAI Chat Completions API with vLLM
 
 vLLM also supports the OpenAI chat completions API format. To follow the process, start the server:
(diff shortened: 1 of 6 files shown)


iboiko-habana:
Merged in https://github.com/vllm-project/vllm-gaudi/pull/1674

PR #1640, opened 2026-07-27
Improve MiniMax-M3 MoE decode performance after upstream merge

mkrze:
## Summary

Reduces HBM bandwidth for MiniMax-M3 SwiGLU-OAI MoE during low-concurrency decode by computing only the routed experts instead of every local expert. This is a follow-up optimization after the upstream MoE merge.

## Motivation

The existing dense SwiGLU-OAI path reads **every** local expert's weights for every token, even though each token routes to only `top_k` experts. At small token counts (single-stream / low-concurrency decode), the number of *distinct* experts the batch touches is at most `tokens * K`, which is often far below `E_local`. Reading all expert weights from HBM is the dominant single-stream decode cost.

## Changes

- **New expert-gather decode path** (`_gather_swigluoai_moe`): gathers only the `G = min(E_local, tokens * K)` routed experts via `index_select`, streaming up to `E_local / G` fewer expert-weight bytes per token. It is:
  - **Bit-exact** with the dense path — reproduces the same fp32 SwiGLU-OAI math and the same per-chunk accumulation/rounding order (ascending expert ids, absent experts contribute exact `+0.0`).
  - **Static-shaped** — `G` depends only on the bucketed token count, so decode-bucket graphs compile once during warmup with no per-step recompile and no data-dependent overflow branch (no token is ever dropped).
- **Dispatcher** (`_swigluoai_moe`): routes to the gather path at low token counts and falls back to the dense path otherwise; wired into both `HPUUnquantizedFusedMoEMethod` forward paths.
- **New env vars** (in `vllm_gaudi/envs.py`, documented in `docs/configuration/env_variables.md`):
  - `VLLM_MINIMAX_M3_MOE_DECODE_GATHER` (default `true`) — enable/disable the gather path.
  - `VLLM_MINIMAX_M3_MOE_GATHER_MAX_TOKENS` (default `16`) — max batch size for the gather path.
  - `VLLM_MINIMAX_M3_MOE_TOKEN_TILE` (default `512`) — now sourced through `vllm_gaudi.envs` instead of a raw `os.environ` read.
- **Tests**: extended `tests/unit_tests/ops/test_hpu_fused_moe.py` to cover the gather path, including bit-exactness against the dense path.

### DIFF
--- a/docs/configuration/env_variables.md
+++ b/docs/configuration/env_variables.md
@@ -26,6 +26,9 @@ This document lists the supported diagnostic and profiling, as well as performan
 | `VLLM_ROW_PARALLEL_CHUNKS`   | Number of chunks to split input into for pipelining matmul with all-reduce in RowParallelLinear layers. Setting to a value greater than 1 enables chunking. See [Row-Parallel Chunking](../features/row_parallel_chunking.md). | `1` (disabled) |
 | `VLLM_ROW_PARALLEL_CHUNK_THRESHOLD` | Minimum number of tokens required to activate row-parallel chunking. Inputs below this threshold use the standard non-chunked path. | `8192` |
 | `VLLM_PROMPT_BS_BUCKET_MAX`  | Sets prefill batch size | `1` |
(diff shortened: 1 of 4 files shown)


adobrzyn (approved):
lgtm

iboiko-habana:
PR is merged in https://github.com/vllm-project/vllm-gaudi/pull/1673

PR #1641, opened 2026-07-27
[FIX_FOR_VLLM_CUSTOM=439f336212227833e126526d3c5f3ef3968dfbf5] Restore is_interleaved helper removed upstream

pawel-olejniczak:
## Root cause
Upstream vLLM PR #49803 (commit 26d725c334) removed the `is_interleaved`
helper from `vllm.transformers_utils.config`, inlining the check at its former
call sites. The HPU model runner still imported that symbol at module load, so
`EngineCore` / worker init crashed with
`ImportError: cannot import name 'is_interleaved' from 'vllm.transformers_utils.config'`
across run_unit_tests, data_parallel, pd_disaggregate and ~50 e2e
load/generate jobs.

## Upstream PR
https://github.com/vllm-project/vllm/pull/49803
Removed `is_interleaved` from `vllm.transformers_utils.config` and inlined the
layer-type check at its call sites.

## Fix
Define a local `is_interleaved(config)` helper in
`vllm_gaudi/v1/worker/hpu_model_runner.py` (identical behaviour to the removed
upstream helper: True when the text config declares more than one distinct
layer type) and drop the now-dead top-level import.

## Bug 2: set use_replayssm on HPUMambaMixer2

- **State machine id**: granite_4_h_hpumambamixer2_use_replayssm_missing
- **Commit**: c219434891d11dfdeb4760c3ea0eda8b8364a21d
- **vllm pin (this fix)**: 439f336212227833e126526d3c5f3ef3968dfbf5
- **Root cause**: Upstream vLLM PR #48018 (ReplaySSM) added `self.use_replayssm` / `self.replayssm_buffer_len` to `MambaMixer2.__init__`, read by the inherited `get_state_shape()` / `get_state_dtype()` during EngineCore init. `HPUMambaMixer2` overrides `__init__` and never set them, so Granite-4-H crashed with `AttributeError: 'HPUMambaMixer2' object has no attribute 'use_replayssm'`.
- **Upstream**: https://github.com/vllm-project/vllm/pull/48018
- **Fix**: Set `use_replayssm=False` and `replayssm_buffer_len=None` in the HPU `__init__` (HPU has no replay decode kernel; conv+SSM path is non-replay).
- **Verification**: HPU re-verify of full stack PASS against vllm@439f336212227833e126526d3c5f3ef3968dfbf5 (granite-4.0-h-small e2e load/generate; both fixes exercised, EngineCore init clean, coherent generations).

### DIFF
--- a/vllm_gaudi/ops/hpu_mamba_mixer2.py
+++ b/vllm_gaudi/ops/hpu_mamba_mixer2.py
@@ -4,6 +4,8 @@
 # Added by the IBM Team, 2024
 # Adapted from https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/mamba/mamba_mixer2.py
 
+import logging
+
 import torch
 import torch.nn.functional as F
 from torch import nn
@@ -42,6 +44,8 @@ from vllm_gaudi.ops.granite_causal_conv1d import (
 )
 from vllm_gaudi.ops.ssd_combined import hpu_mamba_chunk_scan_combined_varlen
 from vllm_gaudi.ops.ops_selector import get_selective_state_update_impl
+
+logger = logging.getLogger(__name__)
 
 
 # Adapted from vllm.model_executor.layers.mamba.mamba_mixer2.Mixer2RMSNormGated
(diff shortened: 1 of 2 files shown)


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -132,6 +131,27 @@
...
+        config: A ``PretrainedConfig`` (or any config exposing
+            ``get_text_config``).
+
+    Returns:
+        True if the text config declares more than one distinct layer type.
+    """
+    text_config = config.get_text_config()
+    if layer_types := getattr(text_config, "layer_types", None):
+        return len(set(layer_types)) > 1
+    return False
```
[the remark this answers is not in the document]
pawel-olejniczak: This helper is a verbatim restore of the upstream implementation that PR #49803 (`26d725c334`) removed from `vllm.transformers_utils.config` (`config.get_text_config()` was the original body), so the behavior at these three call sites is unchanged from what vLLM shipped before the removal.

The `AttributeError` concern does not apply here: `get_text_config()` is a base-class method on `transformers.PretrainedConfig`, so every config inherits it, and on already-extracted text configs it returns `self` (a no-op). The call sites pass `vllm_config.model_config.hf_text_config`, which is typed `PretrainedConfig` and is itself produced by `get_hf_text_config()` (which calls `get_text_config()`), so it always implements the method. Adding a `getattr`/fallback branch would be unreachable dead code, so I'm keeping the restore byte-for-byte identical to upstream. HPU verification against `vllm@439f3362` passed with this code.

vllm_gaudi/ops/hpu_mamba_mixer2.py
```diff
@@ -43,6 +45,8 @@
 from vllm_gaudi.ops.ssd_combined import hpu_mamba_chunk_scan_combined_varlen
 from vllm_gaudi.ops.ops_selector import get_selective_state_update_impl
 
+logger = logging.getLogger(__name__)
```
adobrzyn, line 48: Can we stay consistent with the rest of the repo?

```
from vllm_gaudi.extension.logger import logger as init_logger

logger = init_logger()
```

pawel-olejniczak: Thanks for pointing that out. I'll fix it in the next PR.

iboiko-habana (approved):

PR #1651, opened 2026-07-27
Fix Kimi-K2.5/K2.6 multimodal warmup corruption and dummy-input crash

yeonsily:
Two independent HPU warmup bugs made the Kimi-K2.5/K2.6 vision path unusable unless warmup was disabled entirely (VLLM_SKIP_WARMUP=true):

Corrupted vision RoPE table. Rope2DPosEmbRepeated.get_freqs_cis materializes its freqs_cis buffer lazily on first call and caches it behind a hasattr guard. In graph mode that first call happens during warmup inside PT_COMPILE_ONLY_MODE, where Synapse compiles the recipe but never executes the kernels, so the buffer is filled with uninitialized garbage. The hasattr guard then keeps that garbage forever, so every real image forward uses a corrupt RoPE table and emits repeated token-0 ("!") output. Skip caching while in compile-only mode so the first genuinely-executed forward computes and caches the real (float32, on-device) values. The vision tower runs eager (not hpu-graph-wrapped, compile_mm_encoder off) and its per-shape recipes are unchanged, so this adds no runtime recompilation and does not alter numerics.

Dummy-input count mismatch. Upstream KimiK25DummyInputsBuilder emits one media token per requested item in get_dummy_text but returns a single dummy chunk from get_dummy_mm_data. When limit_per_prompt['vision_chunk'] > 1, KimiK25Processor pops a per-chunk token count for each media token and runs dry, raising "IndexError: pop from empty list" and failing engine init. Replicate the dummy chunk to match the media-token count. Only warmup dummy generation is affected; real requests never touch this path.

Both fixes are scoped to Kimi: the vision-tower patches mutate only vllm.model_executor.models.kimi_k25_vit (whose sole importer is kimi_k25.py), and the processor re-registration targets only KimiK25ForConditionalGeneration.

### DIFF
--- a/vllm_gaudi/models/__init__.py
+++ b/vllm_gaudi/models/__init__.py
@@ -69,3 +69,4 @@ def register_model():
     import vllm_gaudi.models.qwen3_next  # noqa: F401
     import vllm_gaudi.models.qwen3_5  # noqa: F401
     import vllm_gaudi.models.kimi_k25_vit  # noqa: F401
+    import vllm_gaudi.models.kimi_k25  # noqa: F401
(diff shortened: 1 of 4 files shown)


adobrzyn (approved):
lgtm

PR #1654, opened 2026-07-27
Fix Gemma4 E-series PLE compile time on HPU with dynamic indexing (0.26)

jiminha:
Gemma4 E-series models (e4b, e2b) use Per-Layer Embedding (PLE) which causes slow compile times on HPU. The upstream implementation uses Python int indexing per_layer_inputs[:, layer_idx, :] which causes torch.compile to create a separate graph for each layer.

This fix uses torch.index_select with pre-registered layer_idx_tensor buffers, making the indexing dynamic and allowing graph reuse across layers.

Changes:

Add HpuGemma4Model with dynamic PLE indexing in forward()
Add _hpu_run_decoder_layers for fast-prefill path
Register layer_idx_tensor buffers on decoder layers during init
Register HpuGemma4ForConditionalGeneration via ModelRegistry
Results on Gemma-4-E4B-it (max_model_len=131072, max_num_seqs=45):

Total warmup: ~22 mins (was ~40 mins) - ~2x improvement
Prompt compile: ~3 mins (was ~12 mins)
Decode compile: ~18 mins (was ~28 mins)


vllm_gaudi/models/gemma4.py
```diff
@@ -0,0 +1,192 @@
...
+            layer.register_buffer(
+                "layer_idx_tensor",
+                torch.tensor([layer_idx], dtype=torch.long),
+                persistent=False,
+            )
+
+    # Patch _run_decoder_layers for fast-prefill path (self_decoder/cross_decoder)
+    from vllm.model_executor.models import gemma4 as upstream_gemma4_module
+    upstream_gemma4_module._run_decoder_layers = _hpu_run_decoder_layers
+    upstream_gemma4_module.__dict__["_run_decoder_layers"] = _hpu_run_decoder_layers
```
adobrzyn: Correct me if I'm wrong but doesn't those two lines do the same thing? I believe second line can be dropped

`
    upstream_gemma4_module._run_decoder_layers = _hpu_run_decoder_layers
    upstream_gemma4_module.__dict__["_run_decoder_layers"] = _hpu_run_decoder_layers
`

jiminha: yes, 2nd line seems redundant,, let me test and will update.

jiminha: Fixed!

iboiko-habana (approved):

PR #1665, opened 2026-07-29
Add item-count axis to multimodal warmup resolutions

libinta:
Builds on the explicit-resolution warmup already in v0.26.0 (#1651/#1652/ #1663). That warms a single item count per resolution, so a native-resolution tower that flattens N items onto one patch axis (Qwen3.5/3.6, Kimi) recompiles at runtime for the real per-request count. This adds the count axis:

- vision.py: VLLM_MULTIMODAL_RESOLUTIONS now parses WxHxN (pin count N) and WxHxN-M (warm the count range [N,M]); WxH keeps the old meaning (warm at the limit-mm ceiling). qwen3_5/kimi_k25/gemma4 buckets set to [] so warmup skips the guessed aspect-ratio grid and compiles only the operator's resolutions (the guessed shapes match no real request here and only lengthen warmup).

- hpu_model_runner.py: warmup loop expands each resolution over counts_to_warm; _get_mm_dummy_batch takes count_override; _resolve_mm_limit_options falls back from a native modality (Kimi 'vision_chunk') to the generic 'image'/'video' limit-mm key so the per-prompt count is honored. Guessed shapes stay count-1, matching prior behavior. A prefix-caching warning nudges operators toward the range form when only a single count is pinned. 

Validated: qwen3.5 count-20 warmup matches real (no recompile); qwen2.5-vl guessed warmup unchanged; Kimi 864x480 count-20 real matches warmup, no ceiling violation. Unit tests: 22 passed.


docs/features/warmup.md
```diff
@@ -149,16 +149,30 @@ If warm-up is globally skipped, these logs do not appear.
...
+
+```text
+VLLM_MULTIMODAL_RESOLUTIONS=1024x768,768x1024x2,864x480x1-20
+```
+
 The `width`/`height` given via `--limit-mm-per-prompt` (e.g. `--limit-mm-per-prompt '{"image": {"count": 1, "width": 1024, "height": 768}}'`) are unioned with `VLLM_MULTIMODAL_RESOLUTIONS`. Each resolution is built as a raw solid-color image and fed through the model's own processor (`smart_resize`/`navit_resize`/...), rather than a precomputed grid, so the compiled graph matches exactly what a real request at that resolution produces. This is supported for the `image` modality and for Kimi-K2.5/K2.6's `vision_chunk` modality.
 
 ```{.}
-INFO 07-27 23:22:27 hpu_model_runner.py:5845] Using explicit multimodal warmup resolutions (WxH): [(864, 480), (1024, 768), (768, 1024)]
+INFO 07-27 23:22:27 hpu_model_runner.py:5890] Using explicit multimodal warmup resolutions (width, height, count[None=max|N=pin|(lo,hi)=range]): [(864, 480, None), (1024, 768, 2), (768, 1024, (1, 20))]
```
adobrzyn, line 178: NIT: for this log we are using logger.debug not info. We should either fix log or change level in code - this might be misleading if someone uses different logging level

vllm_gaudi/extension/bucketing/vision.py
```diff
@@ -44,28 +44,20 @@
...
         'buckets': [1600, 2048, 3072, 6144, 8192, 131076]
     },
+    # No guessed-bucket warmup; configure via VLLM_MULTIMODAL_RESOLUTIONS /
+    # limit-mm WxH. Entry still required: absence falls back to batch-based
+    # [1,2,4,8].
     'qwen3_5': {
         'is_batch_based': False,
-        # patches per image
-        'buckets': [196, 256, 441, 480, 576, 900, 1156]
+        'buckets': []
```
adobrzyn, line 60: We r emptying buckets for qwen, kimi and gemma so these models now get no vision warmup out of the box. With empty buckets and no VLLM_MULTIMODAL_RESOLUTIONS set, warmup lists is empty and every resolution recompiles on the first real request. Doesn't it?

Also, the new comment says absence "falls back to batch-based [1,2,4,8]". That's not what happens I believe? The entry is still present (just empty), so _get_multimodal_config returns it and there's no fallback path. 

If the intent is "operators must always set VLLM_MULTIMODAL_RESOLUTIONS for these models now," that's fine, but let's make it explicit and fix the comment. Otherwise we're silently regressing three models. Please explain it if I'm wrong

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -5867,68 +5899,90 @@ def warmup_multimodal_graphs(self, buckets):
...
-                        dummy_encoder_outputs,
-                        expected_num_items=candidates[idx],
-                    )
-                    self.graphed_buckets.add(candidates[idx])
-                self.log_warmup_multimodal(phase, idx, len(candidates), candidates[idx] if is_batch_based else 1, 0,
-                                           width, height)
+                    width, height, count_spec = candidates[idx]
+                    if isinstance(count_spec, tuple):
+                        lo, hi = count_spec
+                        counts_to_warm = list(range(lo, min(hi, user_max_count) + 1))
```
adobrzyn, line 6028: This warms one full graph per count per resolution. So example 864x480x1-20 with --limit-mm count=20 compiles 20 graphs for that single resolution, and multiple ranged entries multiply on top of that. Warmup time (and memory) can blow up fast if an operator isn't aware. It's root-controlled and off by default, so not a blocker or anything but can we add a short caution in warmup.md that a wide range = one compile per count? Just so nobody sets 1-50 and wonders why warmup takes forever :D

vllm_gaudi/extension/bucketing/vision.py
libinta: For customer with image requirements, we ask customer to use --limit-mm-per-prompt '{"image": {"count": 20, "width": 864, "height": 480}} or VLLM_MULTIMODAL_RESOLUTIONS for kimi, that will warmup 480p. since the buckets we put before it's guessed buckets anyway. it won't match with customer required anyway. Also as these are new models, it's not regression.  But we can add more in doc. Empty will make it not batch based, your understanding is right.

docs/features/warmup.md
libinta: change to info

vllm_gaudi/v1/worker/hpu_model_runner.py
libinta: We can add a comment in env_variables.md and warmup.md to remind user that warmup the range can exhaust the memory and possible time.
test shows time difference between warmup 1-20 vs only 20
| Model    | RANGE=1 (1–20) | RANGE=0 (count-20) | Δ warmup time   |
| -------- | -------------- | ------------------ | --------------- |
| Kimi     | 2111s          | 2125s              | +14s (noise)    |
| gemma4   | 507s           | 540s               | +33s (noise)    |
| qwen3.5  | 1552s          | 1566s              | +14s (noise)    |


From implementation the warmup with range is needed.
Schedule can schedule fewer than the request's total for three reasons:
encoder cache hit — check_and_update_cache skips already-computed images
encoder compute budget — can_allocate breaks the loop when encoder_compute_budget is exhausted
client simply sent < 20

iboiko-habana (approved):

PR #1694, opened 2026-08-04
[CI] Pin public UBI base image by digest

adobrzyn:
> **Draft** — opening for review of approach before marking ready.

## What

Pins the one **public** container base image the build depends on — `registry.access.redhat.com/ubi9/ubi` — by `sha256` digest instead of the mutable tag, in the two RHEL/UBI Dockerfiles:

- `.cd/Dockerfile.rhel.ubi.vllm`
- `.cd/Dockerfile.rhel.ubi.vllm.no-torchaudio`

This addresses the OpenSSF Scorecard **Pinned-Dependencies** check (container-image portion) for these files.

Uses the **tag + digest** form so the image is both pinned *and* maintainable:

```dockerfile
FROM registry.access.redhat.com/ubi9/ubi:9.6@sha256:dec374e0…eef8bdbc AS gaudi-base
```

- The **digest** is what Scorecard requires (immutable).
- The **`:9.6` tag** lets Dependabot's `docker` ecosystem track the image and open PRs that bump the digest when Red Hat ships a new `9.6` build. A `docker` entry (`directory: /.cd`) is added to `.github/dependabot.yml` for this.

The digest resolves the `ubi9/ubi:9.6` multi-arch manifest list (verified against Red Hat's registry, resolvable by-digest).

## Scope / rationale

Intentionally limited to the **public** UBI image, which is a genuine external supply-chain surface.

**Not** pinned (deliberately):
- **`vault.habana.ai` Gaudi base images** — first-party Intel registry, and their `FROM` lines are fully `ARG`-parameterized (`VERSION` / `REVISION` / `PT_VERSION`) so CI can sweep SynapseAI/PyTorch versions. Dependabot can't resolve a concrete image from those and skips them automatically; pinning by digest would freeze the version-sweep matrix.
- **pip commands** (66 Scorecard warnings) — hash-pinning every `pip install` across build scripts and version-sweep Dockerfiles is brittle and fights the same parameterization; not appropriate for a hardware plugin tracking upstream vLLM/torch.

## Note

Because the base is pinned by digest, keep the `:9.6` tag aligned with `OS_VERSION` (still used for the Habana RPM repo path). A comment on each `FROM` documents this. With the Dependabot `docker` entry, digest bumps are automated going forward.

---
This PR was prepared with the assistance of an AI coding tool.

### DIFF
--- a/.cd/Dockerfile.rhel.ubi.vllm
+++ b/.cd/Dockerfile.rhel.ubi.vllm
@@ -24,7 +24,10 @@ ARG VLLM_GAUDI_REPO="https://github.com/vllm-project/vllm-gaudi.git"
 # ============================================================================
 # Stage 1: gaudi-base - Base system setup with Habana drivers
 # ============================================================================
-FROM registry.access.redhat.com/ubi9/ubi:${OS_VERSION} AS gaudi-base
+# Pin the public UBI base image by digest (OpenSSF Scorecard Pinned-Dependencies).
+# Digest is for registry.access.redhat.com/ubi9/ubi:9.6 (multi-arch manifest list).
+# When bumping OS_VERSION, update this digest to the matching ubi9/ubi:<ver> tag.
+FROM registry.access.redhat.com/ubi9/ubi@sha256:dec374e05cc13ebbc0975c9f521f3db6942d27f8ccdf06b180160490eef8bdbc AS gaudi-base
 
 # Re-declare global ARGs to use them in this stage
 ARG ARTIFACTORY_URL
--- a/.cd/Dockerfile.rhel.ubi.vllm
+++ b/.cd/Dockerfile.rhel.ubi.vllm
@@ -25,9 +25,10 @@ ARG VLLM_GAUDI_REPO="https://github.com/vllm-project/vllm-gaudi.git"
 # Stage 1: gaudi-base - Base system setup with Habana drivers
 # ============================================================================
 # Pin the public UBI base image by digest (OpenSSF Scorecard Pinned-Dependencies).
-# Digest is for registry.access.redhat.com/ubi9/ubi:9.6 (multi-arch manifest list).
-# When bumping OS_VERSION, update this digest to the matching ubi9/ubi:<ver> tag.
-FROM registry.access.redhat.com/ubi9/ubi@sha256:dec374e05cc13ebbc0975c9f521f3db6942d27f8ccdf06b180160490eef8bdbc AS gaudi-base
+# Tag + digest form: the :9.6 tag lets Dependabot's docker ecosystem track and
+# bump the digest; the digest is what Scorecard requires. Keep the tag aligned
+# with OS_VERSION (used below for the Habana RPM repo path).
+FROM registry.access.redhat.com/ubi9/ubi:9.6@sha256:dec374e05cc13ebbc0975c9f521f3db6942d27f8ccdf06b180160490eef8bdbc AS gaudi-base
 
 # Re-declare global ARGs to use them in this stage
 ARG ARTIFACTORY_URL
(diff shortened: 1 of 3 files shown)


PatrykWo:
Thanks for tackling the Scorecard Pinned-Dependencies check — the approach (tag + digest form + Dependabot `docker` entry) is a good one, and the scope reasoning checks out: the other `.cd/` Dockerfiles, `docker-compose.yml`, and the two Dockerfiles outside `.cd/` (`.github/Dockerfile.ci`, `tests/pytorch_ci_hud_benchmark/Dockerfile.hpu`) all use fully ARG-parameterized `vault.habana.ai` bases, so there's nothing else public to pin.

Two things I'd like to resolve before this goes ready:

**1. Hardcoded `:9.6` is now decoupled from `OS_VERSION`.**
`OS_VERSION` still drives `HABANA_RPM_REPO_PATH="rhel/9/${OS_VERSION}"` (and there's an explicit 9.4 build path per the `PT_PACKAGE_NAME_NON_DEFAULT_PYTHON_SUBSTRING` comment). After this change, `--build-arg OS_VERSION=9.4` gives a UBI **9.6** base with **9.4** Habana repos — a silent mismatch that used to be impossible. Since a digest can't be ARG-parameterized without failing Scorecard, this is a real trade-off. Could we either add a build-time guard that fails when `OS_VERSION != 9.6`, or explicitly document that overriding `OS_VERSION` on these two files is no longer supported?

**2. Dependabot will likely propose tag upgrades, not just digest bumps.**
The `docker` ecosystem treats `:9.6` as a version and will open `version-update` PRs to newer tags (9.7, 10.x), not only digest updates. Merging one would move the OS base off 9.6 while `OS_VERSION`/repo path stay at 9.6 — the same mismatch as above, disguised as a routine bump. Suggest constraining the new entry to digest-only:

```yaml
    ignore:
      - dependency-name: "registry.access.redhat.com/ubi9/ubi"
        update-types:
          - version-update:semver-major
          - version-update:semver-minor
```

Minor notes:
- The same digest is duplicated across both Dockerfiles; Dependabot handles both, but manual edits risk drift — just something to keep in mind.
- The digest can't be verified statically. Since this is still a draft, please make sure CI (or a local `docker build`) has actually built one of these images on this branch so a bad/arch-mismatched digest is caught here rather than downstream.

Overall this is close — mainly the `OS_VERSION` decoupling and the Dependabot `ignore` rule. Nice work.

PatrykWo (changes requested):
Hardcoding :9.6 decouples the base image from OS_VERSION and silently breaks non-default builds
The FROM changed from ubi:${OS_VERSION} to a hardcoded ubi:9.6@sha256:..., but OS_VERSION is still live and drives other build logic in the same file:


Dockerfile.rhel.ubi.vllm
Lines 16-16
ARG HABANA_RPM_REPO_PATH="rhel/9/${OS_VERSION}"
The file explicitly documents a non-default 9.4 build path:


Dockerfile.rhel.ubi.vllm
Lines 13-14
for internal use - to support a different package name for RHEL 9.4 with Python 3.12
ARG PT_PACKAGE_NAME_NON_DEFAULT_PYTHON_SUBSTRING=
So docker build --build-arg OS_VERSION=9.4 ... now produces a UBI 9.6 base image while pointing the Habana RPM repo at rhel/9/9.4 — a silent OS/repo mismatch that used to be impossible. The comment says "keep the tag aligned with OS_VERSION," but nothing enforces it; the two are now independent sources of truth that can drift. This is an inherent tension (a digest can't be ARG-parameterized without failing Scorecard), so it's a real trade-off worth calling out in the PR: either add a build-time guard that fails if OS_VERSION != 9.6, or explicitly state that overriding OS_VERSION on these two Dockerfiles is no longer supported.

PR #1720, opened 2026-08-10
Fix MiniMax M2 fused MoE import for vLLM API change

NatTuck:
With this small change, Minimax M2.7 runs on Gaudi 2.

It's kind of slow and tool calls seem a bit glitchy, but running is better than crashing.


iboiko-habana:
@NatTuck  thanks for your PR.

Root cause: upstream vLLM PR [#41184](https://github.com/vllm-project/vllm/pull/41184)  removed the FusedMoE class and exposed expert mapping through the module-level fused_moe_make_expert_params_mapping helper. PR [#44941]https://github.com/vllm-project/vllm/pull/44941 later renamed the constructor to FusedMoEFactory.

iboiko-habana (approved):

PR #1721, opened 2026-08-11
ci: adaptive hourly cadence (8h when healthy, 24h when failing)

adobrzyn:
## Summary
Make the "Hourly Commit Check and Tests" cadence adaptive to pipeline health:

- **Last real run passed → run every 8h**
- **Last real run failed → back off to once every 24h**

GitHub cron can't be conditional, so the workflow is scheduled at the fast **8h** cadence (`0 */8 * * *` → 00:00 / 08:00 / 16:00 UTC) and a new lightweight **`gate`** job decides at runtime whether each tick proceeds.

## How the gate decides
- Runs on a cheap `ubuntu-latest` runner, so a "skip" never locks an HPU runner.
- `workflow_dispatch` always proceeds.
- Otherwise it queries recent runs via the API and finds the most recent **real** run — one where `run_unit_tests` actually executed:
  - **passed** → proceed (8h cadence).
  - **failed** → proceed only once **≥23h** have elapsed since that run started (24h back-off).

### Why "most recent real run" and not "most recent run"
A backed-off tick skips all test jobs, and GitHub reports such a run's overall conclusion as a misleading `success`. Keying off the last run where `run_unit_tests` actually ran prevents those skipped ticks from flipping the cadence back to 8h.

### Why elapsed-time and not exact-hour matching
Scheduled events can be delivered late/drifted by GitHub, so 24h spacing is enforced by comparing elapsed time to the last real run rather than matching a specific UTC hour.

## Wiring
- `discover_runner` is gated on `gate.outputs.should_run`; the rest of the pipeline already chains from it, so a skip cascades cleanly.
- `cleanup_docker_image` (`if: always()`) is guarded so it never runs on a skipped tick, where `discover_runner` was skipped and `runner_name` is empty.

## Notes
- Supersedes #1657 (flat once-a-day schedule) — that PR can be closed if this is preferred.
- The workflow display name still reads "Hourly"; left unchanged to avoid breaking references.

### DIFF
--- a/.github/workflows/hourly-ci.yaml
+++ b/.github/workflows/hourly-ci.yaml
@@ -26,6 +26,9 @@ jobs:
       actions: read # Required to query this workflow's previous runs and their jobs
     outputs:
       should_run: ${{ steps.decide.outputs.should_run }}
+      # ISO-8601 start time of the last real run; empty on bootstrap/manual runs.
+      # setup_and_build uses this as the exact commit-diff window.
+      last_started: ${{ steps.decide.outputs.last_started }}
     steps:
       - name: Decide whether to run
         id: decide
@@ -75,6 +78,9 @@ jobs:
             exit 0
           fi
           echo "Last real run: conclusion=$LAST_CONCLUSION started_at=$LAST_STARTED"
+          # Expose the last real run's start so setup_and_build can scope the
+          # commit diff to exactly "since the last time we actually ran".
+          echo "last_started=$LAST_STARTED" >> "$GITHUB_OUTPUT"
 
           if [ "$LAST_CONCLUSION" = "success" ]; then
             echo "Last real run passed -> 8h cadence -> running."
@@ -118,8 +124,8 @@ jobs:
     if: |
       github.event_name == 'workflow_dispatch' ||
       github.ref == 'refs/heads/main'
-    # <-- UPDATED: Now needs 'discover_tests' AND 'discover_runner'
-    needs: [discover_tests, discover_runner]
+    # <-- UPDATED: Now needs 'gate' (for the diff window), 'discover_tests' AND 'discover_runner'
+    needs: [gate, discover_tests, discover_runner]
     # <-- UPDATED: Runs on the specific runner from the discover_runner job
     runs-on: ${{ needs.discover_runner.outputs.runner_name }}
     permissions:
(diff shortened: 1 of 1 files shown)


adobrzyn:
Thanks @copilot — addressed in d8ab7a55:

1. **`actions: read` permission** — Added `permissions: { actions: read }` to the `gate` job so its Actions API calls (workflow runs + run jobs) don't 403 and block the workflow.

2. **`head -n1` + `set -o pipefail` SIGPIPE** — The `run_unit_tests` conclusion is now selected entirely within jq via `first(.jobs[] | select(...) | .conclusion) // empty`, so there's no `head` closing the pipe early. `// empty` also folds the `null` case into an empty string, so the separate `!= "null"` check was dropped.

3. **`run_started_at` vs `created_at`** — The 24h back-off now measures elapsed time from `run_started_at` (falling back to `created_at`), matching the "since the run started" intent.

4. **`cleanup_docker_image` on `discover_runner` failure** — Guard changed to `if: always() && needs.discover_runner.result == 'success'`. This covers both the gate-skipped case (result `skipped`) and the failure case (result `failure`); in either, `runner_name` is empty and the job would otherwise fail to start.

On the commit-diff window comment: after rebasing onto `main` this branch already picked up #1657's change to a **24h** window (`date -u -d "24 hours ago"`). Since 24h is the maximum spacing under this adaptive schedule (the back-off cadence), the window never *misses* commits — in the healthy 8h case it just over-reports slightly, and that step is informational logging only. Happy to compute an exact per-run window if you'd prefer, but it seemed like unnecessary complexity for a log line.

adobrzyn:
Good point — now properly fixed in 69e4e000 rather than left at a fixed window.

The `gate` job already looks up the last real run's start time, so it now exposes it as an output (`last_started`, from `run_started_at`), and `setup_and_build` uses that value directly as the `git log --since` window. So the commit-diff is scoped to *exactly* "since the last time we actually ran" — 8h when healthy, up to 24h during back-off — instead of a hardcoded window. It falls back to 24h only on bootstrap (no prior run) or manual dispatch, where the gate exposes no start time.

(Note: after the rebase onto `main` the value was already 24h, not 4h — #1657 had changed it — but a fixed window still couldn't track the variable cadence, which this resolves.)

iboiko-habana (approved):

PR #1724, opened 2026-08-12
MiniMax-M3: tolerate dropped "<" in invoke tag; list validated model

mkrze:
## Summary

MiniMax-M3 occasionally drops the leading `<` of a tool-call invoke opening tag,
emitting `]<]minimax[>[invoke name="...">` instead of
`]<]minimax[>[<invoke name="...">`. The HPU MiniMax-M3 tool parser detects invoke
tags via the literal marker `INVOKE_START = NS + "<invoke"`, so a dropped `<`
caused the call to be silently missed — no tool call was produced in the
streaming path, and the entire generation was returned as plain content in the
non-streaming path.

This PR makes invoke detection tolerant of the dropped `<`, mirroring the
parser's existing headless-recovery tolerance for dropped *parameter* opening
tags.

## Changes

- **Parser fix** (`vllm_gaudi/entrypoints/openai/tool_parsers/minimax_m3.py`):
  - Add a variant marker `INVOKE_START_NOLT = NS + "invoke"` and a
    `_find_invoke_open()` helper that returns the earliest of the well-formed or
    dropped-`<` opener along with its length (preferring the well-formed form on
    ties).
  - Use it in the streaming path, and accept either opener in the non-streaming
    `_parse_block` / `_parse_one_invoke`.
  - The fix is **position-preserving** (no text rewriting), so streaming offset
    tracking is unaffected, and it **cannot false-positive** on well-formed text
    (a real invoke always has `<` right after the namespace marker). Scope is
    limited to the opening invoke tag; `INVOKE_END` and other tags are untouched.
- **Tests** (`tests/unit_tests/test_minimax_m3_tool_parser.py`): add regression
  tests for the dropped-`<` case in both the non-streaming and streaming paths.
- **Docs** (`docs/getting_started/validated_models.md`): add MiniMax-M3
  (TP=8, BF16, Gaudi 3) to the validated models list.

### DIFF
--- a/vllm_gaudi/entrypoints/openai/tool_parsers/minimax_m3.py
+++ b/vllm_gaudi/entrypoints/openai/tool_parsers/minimax_m3.py
@@ -78,6 +78,11 @@ NS = "]<]minimax[>["
 TOOL_CALL_START = NS + "<tool_call>"
 TOOL_CALL_END = NS + "</tool_call>"
 INVOKE_START = NS + "<invoke"
+# MiniMax-M3 rarely drops the leading "<" of the invoke open tag, emitting
+# "]<]minimax[>[invoke name=..." instead of "]<]minimax[>[<invoke name=...".
+# Tolerate that variant, mirroring the headless-parameter recovery already done
+# for dropped parameter opening tags (see the NS branches below).
+INVOKE_START_NOLT = NS + "invoke"
 INVOKE_END = NS + "</invoke>"
 ELEMENT_START = NS + "<"
 ELEMENT_END_START = NS + "</"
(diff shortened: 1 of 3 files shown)


vllm_gaudi/entrypoints/openai/tool_parsers/minimax_m3.py
```diff
@@ -564,11 +588,15 @@ def _tool_props(self, func_name: str, request) -> dict:
 
     def _parse_one_invoke(self, inv_block: str, request) -> ToolCall:
```
adobrzyn, line 609: We're now expressing "an invoke opener is INVOKE_START or INVOKE_START_NOLT" in three spots: _find_invoke_open, _parse_one_invoke, and I believe _parse_block each a bit differently. Have you considered pulling it into a small helper function, so we will have set of openers in one place? Given we already do headless recovery for dropped param tags, I could see another dropped-< variant showing up later, and I'd rather not have to update three call sites in sync. Non-blocking but worth considering

adobrzyn: Thinking something like that:

```
def helper:
 for opener in INVOKE_OPENERS:
  if s.startswith(opener):
   return len(opener)
  return None
```

Tho, worth checking if that would work :D

mkrze: Done in 249fd183. Pulled the openers into a single INVOKE_OPENERS tuple with a _match_invoke_opener() helper, and routed _find_invoke_open, _parse_one_invoke, and _parse_block through it. A future dropped-tag variant is now a one-line add to INVOKE_OPENERS.

iboiko-habana (approved):

PR #1738, opened 2026-08-18
Enable flatten_input for model type "qwen3_5", and "qwen3_5_text" under language-model-only

### LINKED ISSUE #1735: [Bug] KV cache index_copy_ slot mapping mismatch in eager mode with GDN models on v0.26.0
## Bug Description

vllm-gaudi v0.26.0 fails with an `index_copy_` IndexError in the GDN state cache path when running in eager mode (`--enforce-eager`). Even bypassing graph compilation, the GDN KV cache write path has a regression where `index_copy_` receives 128 slot indices but the cache tensor expects 16384 rows per slot. The same configuration works correctly on v0.24.0.

## Environment

- **Image**: `vault.habana.ai/gaudi-docker/1.24.1/ubuntu24.04/habanalabs/vllm-0.26.0-ptupstream-2.11.0:latest`
- **vLLM version**: v0.26.0
- **Gaudi software**: v1.24.1 (PyTorch 2.11)
- **Model**: `bottlecapai/ThinkingCap-Qwen3.6-27B-FP8` (hybrid: FullAttention + GDN layers)
- **Configuration**: TP=4, max-model-len=262144, kv-cache-dtype=fp8_inc, VLLM_SKIP_WARMUP=true, --enforce-eager, no --speculative-config, no VLLM_COMPACT_GDN

## Reproduction

```bash
docker run -d \
  --name qwen36-fp8 \
  --runtime=habana \
  -e HABANA_VISIBLE_DEVICES=all \
  -e VLLM_SKIP_WARMUP=true \
  -v /local/hf_cache:/root/.cache/huggingface \
  --cap-add=sys_nice \
  -p 8001:8001 \
  --ipc host \
  --entrypoint bash \
  vault.habana.ai/gaudi-docker/1.24.1/ubuntu24.04/habanalabs/vllm-0.26.0-ptupstream-2.11.0:latest \
  -c 'vllm serve bottlecapai/ThinkingCap-Qwen3.6-27B-FP8 --served-model-name qwen36-fp8 --host 0.0.0.0 --port 8001 --tensor-parallel-size 4 --trust-remote-code --max-model-len 262144 --enable-auto-tool-choice --tool-call-parser qwen3_coder --reasoning-parser qwen3 --kv-cache-dtype fp8_inc --max-num-batched-tokens 8192 --max-num-seqs 256 --language-model-only --enforce-eager'

# Wait for startup, then send a request:
curl -X POST http://localhost:8001/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"qwen36-fp8","messages":[{"role":"user","content":"Hello"}],"max_tokens":50}'
```

## Error

```
(Worker_TP3 pid=890) ERROR 08-13 17:54:57 [multiproc_executor.py:1007]   File "/usr/local/lib/python3.12/dist-packages/vllm_gaudi/extension/utils.py", line 75, in forward
(Worker_TP3 pid=890) ERROR 08-13 17:54:57 [multiproc_executor.py:1007]     cache.index_copy_(0, slot_mapping, input)
(Worker_TP3 pid=890) ERROR 08-13 17:54:57 [multiproc_executor.py:1007] IndexError: index_copy_(): Number of indices (128) should be equal to source.size(dim) (16384)
```

The error occurs in `vllm_gaudi/extension/utils.py` line 75, in the GDN state cache `forward()` method. The `slot_mapping` tensor has 128 entries but the `input` tensor (cache values to write) has 16384 entries along dimension 0 — a shape mismatch of 128x.

## Root Cause Analysis

This appears to be a v0.26.0-specific regression in the GDN state cache path. The `index_copy_` operation at `utils.py:75` receives a `slot_mapping` with 128 indices (matching the number of KV cache slots for the request) but the `input` tensor has 16384 rows (likely the full GDN state cache size per layer). The slot mapping and input tensor shapes are not aligned in v0.26.0's GDN cache code.

On v0.24.0, the same code path works correctly, suggesting the regression was introduced during the v0.26.0 realignment with upstream vLLM API drift.

## Workaround

Use v0.24.0 (`vllm-0.24.0-ptupstream-2.11.0` image) which works correctly with this model in both eager and graph modes.

## Related

- #1284 — Original feature request for Qwen3.5 GDN support on Gaudi
- This bug is independent of the graph compilation failure (Bug 1), but both affect the same model class on v0.26.0

shepark:
## Summary
Fixes #1735. Serving Qwen3.5/3.6 (`Qwen3_5ForConditionalGeneration`) with
`--language-model-only` crashes in decode:
`IndexError: index_copy_(): Number of indices (128) should be equal to source.size(dim) (16384)`.

## Root cause
`--language-model-only` makes `supports_multimodal_inputs()` return False, so
the runner passes raw 2-D `input_ids` instead of 2-D `inputs_embeds`. Without
`flatten_input` the model emits 3-D hidden_states → K/V reshape to `batch*seq`
(16384) vs `slot_mapping` (128) → `index_copy_` fails. `flatten_input` only
matched `qwen3_5_moe`, but this checkpoint reports `qwen3_5` / `qwen3_5_text`.
Model/config-specific, not hardware-specific.

## Changes
- `features.py`: add `qwen3_5`, `qwen3_5_text` to `flatten_input`.
- `hpu_model_runner.py`: add them to `is_qwen_moe`.
- `qwen3_5_moe` left untouched.

## Test
FP8, TP=4, `--enforce-eager`, `--language-model-only`: crashes before, serves
after. Without `--language-model-only` still works.

Fixes #1735


vllm_gaudi/extension/features.py
```diff
@@ -106,8 +106,10 @@ def get_features():
         Value('dynamic_shapes_compilation', True, env_var='VLLM_T_COMPILE_DYNAMIC_SHAPES', env_var_type=boolean),
         Value('fullgraph_compilation', False, env_var='VLLM_T_COMPILE_FULLGRAPH', env_var_type=boolean),
         Value('scale_adjustment', True, env_var='VLLM_SCALE_ADJUSTMENT', env_var_type=boolean),
-        Value('flatten_input',
-              Any(ModelType('qwen3_moe'), ModelType('granitemoe'), ModelType('glm4_moe'), ModelType('gemma4'))),
+        Value(
+            'flatten_input',
+            Any(ModelType('qwen3_moe'), ModelType('qwen3_5'), ModelType('qwen3_5_text'), ModelType('granitemoe'),
+                ModelType('glm4_moe'), ModelType('gemma4'))),
```
adobrzyn, line 112: We're adding qwen3_5/qwen3_5_text to flatten_input but leaving out qwen3_5_moe - which is the one variant that's in is_qwen_moe but not here. Is that intentional (MoE input path differs), or should it flatten too? Same question really applies to qwen3_moe being in but qwen3_5_moe out? Please help me out, too many qwens :D

shepark: @adobrzyn thank you for the review and your question gave me a good opportunity to read codes in detail :)
* `is_qwen_moe`: The name is misleading. It actually gates whether a model is a Qwen3-Next-family hybrid (GDN + full-attention, regardless of MoE) that needs the residual=zeros fix and warmup. Since qwen3_5 and qwen3_5_text are hybrid (though dense), they must remain here to receive the fix. So, maybe.. rename properly this time?, something like `is_qwen3_next_family` to avoid confusion.
* `flatten_input`: qwen3_5_moe was intentionally removed (#1713) due to performance regressions and will not be re-added.
* `qwen3_moe`: Kept as-is for now, as we haven't verified if it suffers from the same regression. This is out of scope for this PR.

iboiko-habana (approved):

adobrzyn (approved):

iboiko-habana (approved):

PR #1744, opened 2026-08-20
[Bugfix][HPU] Add opt-in env to run MM warmup outside PT_COMPILE_ONLY_MODE

libinta:
gemma-4's vision pooler strips padding with a data-dependent gather (pooled_states[valid_mask]) whose output row count is only known by executing the op. In torch.compile mode (PT_HPU_LAZY_MODE=0) multimodal warmup runs inside PT_COMPILE_ONLY_MODE, which traces recipes without executing them, so the gather's output shape is never materialized. That leaves an uninitialised tensor offset and can intermittently crash warmup with "offset_ 0is not matching with the offset of new tensor 18446744069...".
specifically, the code in gemma4_mm.py vllm upstream valid_states = pooled_states[valid_mask]

Running MM warmup outside compile-only executes the vision path for real and avoids the hazard, but costs extra warmup time, so keep the existing inside-compile-only fast path as the default. Set
VLLM_MM_WARMUP_OUTSIDE_COMPILE_ONLY=1 to warm multimodal graphs outside compile-only in torch.compile mode (lazy mode already does this). Default off preserves current behavior.


iboiko-habana (changes requested):
1) mm_warmup_outside = os.environ.get(
            "VLLM_MM_WARMUP_OUTSIDE_COMPILE_ONLY", "0").strip().lower() in ("1", "true")

please re-write it via gaudi_envs usage

```
diff --git a/vllm_gaudi/v1/worker/hpu_model_runner.py b/vllm_gaudi/v1/worker/hpu_model_runner.py
index 4f38b205..355bab7d 100644
--- a/vllm_gaudi/v1/worker/hpu_model_runner.py
+++ b/vllm_gaudi/v1/worker/hpu_model_runner.py
@@ -27,6 +27,7 @@ import torch
 import torch.distributed
 import torch.nn.functional as F
 import torch.nn as nn
+import vllm_gaudi.envs as gaudi_envs
 import vllm_gaudi.extension.environment as environment
 from vllm_gaudi.extension.bucketing.common import HPUBucketingManager
 from vllm_gaudi.extension.defragmentation import OnlineDefragmenter
```

mm_warmup_outside = gaudi_envs.VLLM_MM_WARMUP_OUTSIDE_COMPILE_ONLY

2) VLLM_MM_WARMUP_OUTSIDE_COMPILE_ONLY shoud be added to vllm_gaudi/envs.py
i.e.

```
diff --git a/vllm_gaudi/envs.py b/vllm_gaudi/envs.py
index ef0406fd..c1a1fddf 100644
--- a/vllm_gaudi/envs.py
+++ b/vllm_gaudi/envs.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
     VLLM_MINIMAX_M3_MOE_TOKEN_TILE: int = 512
     VLLM_MINIMAX_M3_MOE_DECODE_GATHER: bool = True
     VLLM_MINIMAX_M3_MOE_GATHER_MAX_TOKENS: int = 16
+    VLLM_MM_WARMUP_OUTSIDE_COMPILE_ONLY: bool = False

 # The begin- and end here are used by the documentation generator
 # to extract the used env vars.
@@ -81,6 +82,11 @@ environment_variables: dict[str, Callable[[], Any]] = {
     lambda: os.environ.get("VLLM_MINIMAX_M3_MOE_DECODE_GATHER", "1").lower() in ("1", "true"),
     "VLLM_MINIMAX_M3_MOE_GATHER_MAX_TOKENS":
     lambda: int(os.environ.get("VLLM_MINIMAX_M3_MOE_GATHER_MAX_TOKENS", "16")),
+
+    # Run multimodal warmup outside PT_COMPILE_ONLY_MODE for models with
+    # data-dependent output shapes that must be materialized during warmup.
+    "VLLM_MM_WARMUP_OUTSIDE_COMPILE_ONLY":
+    lambda: os.environ.get("VLLM_MM_WARMUP_OUTSIDE_COMPILE_ONLY", "false").strip().lower() in ("1", "true"),
 }

 # end-env-vars-definition
```

3) new env var should be described in documentation: docs/configuration/env_variables.md and docs/features/warmup.md
i.e.
```
git diff docs/configuration/env_variables.md
diff --git a/docs/configuration/env_variables.md b/docs/configuration/env_variables.md
index 0e4ecbf8..3326cbab 100644
--- a/docs/configuration/env_variables.md
+++ b/docs/configuration/env_variables.md
@@ -14,6 +14,7 @@ This document lists the supported diagnostic and profiling, as well as performan
 | `VLLM_T_COMPILE_FULLGRAPH`                | Forces the PyTorch compile function to raise an error if any graph breaks happen during compilation. This allows for the easy detection of existing graph breaks, which usually reduce performance.                                                                                                                                                                                                                     | `false`       |
 | `VLLM_T_COMPILE_DYNAMIC_SHAPES`           | Forces PyTorch to compile graphs with disabled dynamic options to use dynamic shapes only when needed.                                                                                                                                                                                                                                                                                                                  | `false`       |
 | `VLLM_FULL_WARMUP`                        | Forces PyTorch to assume that the warm-up phase fully covers all possible tensor sizes, preventing further compilation. If compilation occurs after warm-up, PyTorch will crash (with this message: `Recompilation triggered with skip_guard_eval_unsafe stance. This usually means that you have not warmed up your model with enough inputs such that you can guarantee no more recompilations.`) and must be disabled. | `false`       |
+| `VLLM_MM_WARMUP_OUTSIDE_COMPILE_ONLY`     | Runs multimodal warmup outside `PT_COMPILE_ONLY_MODE`. Enable this for multimodal paths with data-dependent output shapes that must be materialized during warmup, such as Gemma4. Enabling it increases warmup time because recipes are compiled and executed.                                                                                                                                                | `false`       |

 ## Performance Tuning Parameters
```

```
git diff docs/features/warmup.md
diff --git a/docs/features/warmup.md b/docs/features/warmup.md
index b71e8264..18fe7b7d 100644
--- a/docs/features/warmup.md
+++ b/docs/features/warmup.md
@@ -149,6 +149,14 @@ If warm-up is globally skipped, these logs do not appear.

 Native-resolution vision towers (e.g. Gemma4, Kimi-K2.5/K2.6, Qwen2.5/3/3.5-VL) produce a different patch grid per image resolution, unlike batch-based towers (e.g. Gemma3) which only vary by batch size. If a real request's resolution does not match anything warmed at startup, the vision-tower graph recompiles on the critical path.

+In lazy mode, multimodal warm-up runs outside `PT_COMPILE_ONLY_MODE`. In `torch.compile` mode, it runs inside compile-only mode by default so recipes are compiled without being executed. Some multimodal paths have data-dependent output shapes that must be materialized during warm-up. For those models, set `VLLM_MM_WARMUP_OUTSIDE_COMPILE_ONLY=true` to compile and execute multimodal warm-up outside compile-only mode:
+
+```text
+VLLM_MM_WARMUP_OUTSIDE_COMPILE_ONLY=true
+```
+
+This setting increases startup time and applies only to multimodal warm-up in `torch.compile` mode. It has no effect when `VLLM_SKIP_WARMUP=true`.
+
 By default, warm-up derives a spread of aspect-ratio shapes from the model's patch-count buckets (`VLLM_MULTIMODAL_BUCKETS`). Some native-resolution towers (Gemma4, Kimi-K2.5/K2.6, Qwen3.5) ship with **empty** patch-count buckets on purpose — the guessed aspect-ratio shapes match no real request for these models and only lengthen warm-up. For them there is no guessed warm-up at all, so you **must** set `VLLM_MULTIMODAL_RESOLUTIONS` (or `--limit-mm-per-prompt` width/height) to precompile anything; otherwise the vision-tower graph recompiles on the first real request at each resolution (the runner logs a warning at startup when this happens). To guarantee a specific production resolution is precompiled, set `VLLM_MULTIMODAL_RESOLUTIONS` to a comma-separated list of raw pixel resolution entries:

 ```text

iboiko-habana (approved):

PR #1745, opened 2026-08-20
Fix decode-block clamp misdetecting hybrid/SWA models (regression from #1700)

adobrzyn:
## Problem

#1700 clamped non-contiguous PA decode block buckets to `max_blocks * 3` and tried to exclude "hybrid" models via `attn_block_size != block_size`. That signal only fires for non-GDN mamba models (Granite-4.0-H, inflated `cache_config.block_size`). For everything else `attn_block_size == block_size`, so **Qwen3.5** (GDN; HPU kernel at 128-token granularity) and **Gemma4** (interleaved sliding-window; `num_gdn == 0`, `num_mamba_like_layers == 0`) were treated as non-hybrid and clamped.

Their decode `block_list` references legitimately exceed 3× physical (GDN virtual block splitting; Gemma's global-attention layers keep full-context block tables), so runtime lookups miss the warmed buckets and recompile the decode graph — the reported regression.

## Fix

Detect the "leave the decode block range unbounded" set using the real signals available on the runner before warmup:

```python
self.bucketing_manager.unbounded_decode_blocks = (
    self.attn_block_size != self.block_size   # Granite-4.0-H (inflated block_size)
    or self.num_gdn > 0                        # GDN hybrids  → Qwen3.5
    or self.num_mamba_like_layers > 0          # mamba hybrids
    or self.interleaved_sliding_window)        # interleaved SWA → Gemma4
```

Plain dense models (Qwen3-0.6B, Llama) keep the clamp, preserving the original first-decode warmup-OOM fix.

Also renames the bucketing flag `is_hybrid` → `unbounded_decode_blocks` to describe what it controls and avoid confusion with the unrelated vLLM `model_config.is_hybrid`.

## Testing

- `pytest tests/unit_tests/test_bucketing.py` → 40 passed
- yapf + ruff clean
- Hybrid unit tests updated to the renamed kwarg
- **Pending:** HPU validation that Qwen3.5 and Gemma4 no longer recompile decode buckets at runtime, and that dense models still avoid the warmup OOM.

### DIFF
--- a/tests/unit_tests/test_decode_bucket_hybrid.py
+++ b/tests/unit_tests/test_decode_bucket_hybrid.py
@@ -90,15 +90,16 @@ class _MockConfig:
             object.__setattr__(self, k, v)
 
 
-def _make_bucketing_manager(block_size, max_model_len, max_num_seqs, num_hpu_blocks, is_hybrid=True):
+def _make_bucketing_manager(block_size, max_model_len, max_num_seqs, num_hpu_blocks, skip_decode_block_clamp=True):
     """Create a minimally-configured HPUBucketingManager.
 
-    Defaults to is_hybrid=True since this helper is used exclusively for hybrid
-    model scenarios, which keep the decode block range unbounded (not clamped).
+    Defaults to skip_decode_block_clamp=True since this helper is used
+    exclusively for hybrid model scenarios, which keep the full worst-case
+    decode block range (skip the 3x-physical clamp).
     """
     mgr = HPUBucketingManager.__new__(HPUBucketingManager)
     mgr.block_size = block_size
-    mgr.is_hybrid = is_hybrid
+    mgr.skip_decode_block_clamp = skip_decode_block_clamp
     mgr.max_model_len = max_model_len
     mgr.max_num_seqs = max_num_seqs
     mgr.max_num_prefill_seqs = 1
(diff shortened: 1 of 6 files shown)


adobrzyn:
## Validation

Validated on Gaudi3 (single card), V1 engine, **default serving** (prefix caching ON → non-contiguous PA — the path the clamp affects), no `--enforce-eager` so warmup + graph compile actually run. Load driver: 64 concurrent `/v1/completions` sharing a long prefix, 3 rounds (192 requests), `max_tokens=256`, `ignore_eos`, `temperature=0`.

Each config is chosen so the buggy `min(worst_case, 3 × num_hpu_blocks)` clamp is **active** (worst_case > 3 × num_hpu_blocks) — otherwise the clamp is a no-op and fix vs. bug are indistinguishable.

| Model | Detection | num_hpu_blocks | buggy clamp (3×) | decode `block` max_warmup | runtime decode recompiles | OOM |
|---|---|---|---|---|---|---|
| Qwen3-0.6B (dense) | dense → clamp kept | 7593 | 22779 | 22779 | **0** | **0** |
| Qwen3.5-35B-A3B | GDN (30 layers) → unbounded | 1744 | 5232 | 8192 | **0** | **0** |
| google/gemma-4-E4B-it | interleaved SWA → unbounded | 4399 | 13197 | 20480 | **0** | **0** |

**Qwen3-0.6B (regression guard):** still detected as dense, so the clamp stays active (block max = 22779 = 3 × 7593). Warmup completes with **no OOM** → the dense OOM guard from #1700 is preserved.

**Qwen3.5-35B-A3B & gemma-4-E4B-it (the regression this PR fixes):** now correctly flagged `unbounded_decode_blocks` (GDN / interleaved SWA), so decode block buckets are warmed to the full worst case (8192 / 20480). That max **exceeds** the old clamp value (5232 / 13197) — exactly the block range the bug left un-warmed and that produced runtime `"Decode bucket for ... was not prepared. Adding new bucket"` recompiles on these two models. Under concurrent long-shared-prefix decode: **0 recompiles, 0 OOM.**

All three pass.

iboiko-habana (approved):

PR #1753, opened 2026-08-25
Disable Synapse persistent-input reuse for compact-GDN

asinbarajx:
## Summary
With compact-GDN (`VLLM_COMPACT_GDN`), the GDN recurrent-state (conv/ssm) read and write are split across separate `torch.compile` recipes. Synapse's per-graph persistent-input reuse can then reuse a still-live state buffer's memory as intra-graph scratch — in the recipe that only reads the state it looks like a dead intermediate, but another recipe / the next decode step still needs it — corrupting the state and producing NaN output (gsm8k 0.00, all-`!` tokens) on Qwen3.5-A17B FP8 at TP=8.

The reuse is opt-in per input and decided one recipe at a time, so neither Synapse nor the bridge's per-graph analysis can detect this cross-recipe reuse. This disables it on the compact-GDN path via the existing `PT_HPU_ENABLE_SYNAPSE_INPUT_REUSE` flag, set in `set_compile_env_defaults()` (torch.compile-only, no-op in lazy, and it never overwrites a user-set value). It only forgoes an input-DRAM-reuse memory optimization — no compute/latency impact.

## Test plan
Gaudi3, Qwen3.5-A17B FP8, TP=8, `VLLM_COMPACT_GDN`, gsm8k (lm_eval, skip warmup):
- before: gsm8k **0.00** (NaN collapse)
- after: gsm8k **~0.90**, stable across repeats


vllm_gaudi/platform.py
```diff
@@ -454,6 +455,20 @@ def set_compile_env_defaults(cls) -> None:
...
+        # across separate torch.compile recipes. Synapse's per-graph persistent-input
+        # reuse optimization can then reuse a still-live state buffer's memory as
+        # intra-graph scratch (the reading recipe cannot see that another recipe / the
+        # next step still needs it), corrupting the state and producing NaN output.
+        # The optimization is opt-in per input via the bridge; the reuse decision is
+        # made one recipe at a time and cannot detect this cross-recipe reuse, so we
+        # turn it off on this path. Disabling it may slightly raise peak memory (a
+        # persistent input's memory is no longer reused as scratch) but has no
+        # compute/latency impact. Never overwrites a user-provided value.
+        if gaudi_envs.VLLM_COMPACT_GDN:
```
pawel-olejniczak: `set_compile_env_defaults()` runs before anything sets `VLLM_COMPACT_GDN`, so on the auto-detected path this gate is always False.

The only production caller is `check_and_update_config()` at platform.py:146, which vLLM invokes from `VllmConfig.__post_init__`. The only writer of `VLLM_COMPACT_GDN` is `HPUModelRunner.__init__` (hpu_model_runner.py:1397-1403), and the runner is constructed later, in `HPUWorker.init_device()` (hpu_worker.py:166) - in the worker process, so that write is not even visible to the process that evaluated the gate. `PT_HPU_ENABLE_SYNAPSE_INPUT_REUSE` therefore stays unset for exactly the case the docs row you added calls the norm ("Auto-detected per model during engine init", docs/configuration/env_variables.md:54).

The test plan lists `VLLM_COMPACT_GDN` in the launch env, which is the one case where the gate does fire, and all four new unit tests pre-set the variable before calling `set_compile_env_defaults()`, so neither exercises the auto path. Did you reproduce the NaN collapse and the fix without exporting `VLLM_COMPACT_GDN`? If not, the decision likely has to move to where the config is available (derive it from `vllm_config` in `check_and_update_config`) or to the auto-detection site itself - the latter runs after `init_device()`, so it would need a check that the bridge still picks the flag up that late.

vllm_gaudi/platform.py
```diff
@@ -454,6 +455,20 @@ def set_compile_env_defaults(cls) -> None:
...
+            if os.environ.get('PT_HPU_ENABLE_SYNAPSE_INPUT_REUSE') is None:
```
pawel-olejniczak: ruff 0.11.7, the version pinned in `.pre-commit-config.yaml`, reports SIM102 on this nested `if`, and there is no safe autofix, so the hook's `--fix` will not clear it:

```
vllm_gaudi/platform.py:469:9: SIM102 Use a single `if` statement instead of nested `if` statements
```

It is the only ruff error in the tree at this SHA, and the `pre-commit` job gates `pre_merge_hpu_test_build`, so pre-merge stops before the HPU build. Collapse it to `if gaudi_envs.VLLM_COMPACT_GDN and os.environ.get('PT_HPU_ENABLE_SYNAPSE_INPUT_REUSE') is None:`.

vllm_gaudi/ops/hpu_gdn_pytorch.py
```diff
@@ -36,9 +36,18 @@
 _USE_EXACT_SOLVE = os.getenv("VLLM_GDN_EXACT_SOLVE", "0") == "1"
 
 
-@torch._dynamo.disable
 def _preprocess_qk_l2norm(q, k):
-    """L2norm in eager mode — HPU torch.compile miscompiles l2norm."""
+    """L2norm the q/k pair ahead of the chunk pipeline.
+
+    This was previously @torch._dynamo.disable'd to keep l2norm in eager mode.
```
pawel-olejniczak: Dropping `@torch._dynamo.disable` here reverts the accuracy workaround added by "qwen35 compact mode" (052a0d15, #1235: "Work-around accuracy issue introduced by torch compile for few ops in hpu_gdn_pytorch"), and the PR description does not mention the change.

The stated justification also does not hold: the recurrent-state read and write sit in separate recipes regardless of this break. The read is `initial_state = ssm_state[state_indices]` inside `_extract_metadata` (vllm_gaudi/models/qwen3_5.py:114), whose own docstring notes that "Dynamo graph-breaks naturally on `get_forward_context()`" (qwen3_5.py:81), while the write happens later at qwen3_5.py:239. Removing the l2norm break does not merge those two into one recipe.

There is also an open PR doing exactly this removal, #1457, which pairs it with a switch to `torch.ops.hpu.l2_norm` plus a dtype fix and is still marked DO_NOT_MERGE.

Can you split this hunk out so the env-var fix lands on its own, or post the A/B showing the pure-PyTorch `_l2norm_last_dim` is accurate under compile at prefill shapes? If it stays, the docstring should describe what the function does rather than what the decorator used to do.

pawel-olejniczak (commented):
Reviewed at `c8871cc3cf79b5262803e7241cec77d378999327`. The mechanism writeup in the description is clear and the "never overwrite a user value" shape matches the existing defaults in `set_compile_env_defaults()`. Three points inline.

vllm_gaudi/platform.py
asinbarajx: Good catch, and confirmed on HW. I reproduced it *without* exporting `VLLM_COMPACT_GDN`: on the auto-detected path the old gate never fired — the model runner sets `VLLM_COMPACT_GDN` in the worker (`init_device`), after `check_and_update_config` — so stock still collapsed (gsm8k **0.00**).

Moved the decision into `check_and_update_config`, deriving compact-GDN from `vllm_config` (`_compact_gdn_active`: `num_gdn > 0` via `get_num_layers_by_block_type`, `kv_transfer_config is None`, `granitemoehybrid` excluded; an explicit `VLLM_COMPACT_GDN` still wins). The flag is now set at engine construction, before workers spawn. Auto path is clean with the fix: gsm8k **0.90/0.92** (Qwen3.5-A17B FP8, TP=8, limit 100). Pushed in `a4ce255`.

vllm_gaudi/platform.py
asinbarajx: Fixed in `a4ce255`. The logic moved into `_maybe_disable_synapse_input_reuse`, which uses sequential early-returns instead of a nested `if`, so SIM102 no longer applies.

pawel-olejniczak (approved):

iboiko-habana (approved):

PR #1760, opened 2026-08-27
GLM-5.2-FP8 with DSA enablement

jkaniecki:
Enable GLM-5.2 with DSA (DeepSeek Sparse Attention) on HPU

What changed:

`hpu_sparse_attn_indexer.py` (new): per-request Q·K BF16 scoring + torch.topk to select top-2048 KV cache slots per decode step
`hpu_attn.py`: forward_mqa_sparse — gathers top-K latent KV, decompresses, runs masked MLA decode attention
`oot_mla.py`: dispatches to forward_mqa_sparse when use_sparse=True
`deepseek_v2.py`: BF16 indexer cache, HPU-safe Indexer.forward, SparseAttnIndexer dispatch, removed `indexers_proj` weights filter as it's no longer needed for current model version
`platform.py / hpu_model_runner.py`: DSA routing + indexer cache layer index fix
Validated on 8× Gaudi3 (GLM-5.2-FP8, TP=8):

GSM8K 5-shot (1319 samples): 93.6% exact match
AIME 2026 - 100% - on pair with model card's target (99,2 % mean for multiple reruns)

### DIFF
--- a/vllm_gaudi/attention/backends/hpu_attn.py
+++ b/vllm_gaudi/attention/backends/hpu_attn.py
@@ -353,6 +353,39 @@ class HPUMLAImpl(MLACommonImpl[HPUAttentionMetadata], torch.nn.Module):
                                                   kv_lora_rank=self.kv_lora_rank)
         return output
 
+    @torch.compiler.disable
+    def forward_mqa_sparse(self, q, k_cache, attn_metadata, topk_indices):
+        """Sparse MLA decode: attend to only the top-K cache entries."""
+        if isinstance(k_cache, tuple):
+            k_cache = k_cache[0]
+        batch_size = q.shape[0]
+        topk = topk_indices.shape[1]
+
+        # Gather top-K latent KV from cache using physical slot indices
+        flat_idx = topk_indices[:batch_size].reshape(-1)
+        selected = k_cache[flat_idx].view(batch_size, topk, -1)
+
+        # Decompress KV
+        k_c, k_pe = selected.split(
+            [self.kv_lora_rank, self.qk_rope_head_dim], dim=-1)
+        kv_nope = self.kv_b_proj(k_c.reshape(-1, self.kv_lora_rank))[0]
+        kv_nope = kv_nope.view(
+            batch_size, topk, self.num_heads,
+            self.qk_nope_head_dim + self.v_head_dim)
+        k_nope, v = kv_nope.split(
+            [self.qk_nope_head_dim, self.v_head_dim], dim=-1)
+        k_pe_exp = k_pe.unsqueeze(2).expand(-1, -1, self.num_heads, -1)
+        key = torch.cat([k_nope, k_pe_exp], dim=-1)
+
+        # q: [B, H, D] -> [B, H, 1, D]
+        # key: [B, T, H, D] -> [B, H, T, D]
+        key = key.permute(0, 2, 1, 3)
+        v = v.permute(0, 2, 1, 3)
+        attn = torch.matmul(q.unsqueeze(2), key.transpose(-1, -2))
+        attn = torch.softmax(attn * self.scale, dim=-1)
+        out = torch.matmul(attn, v).squeeze(2)
+        return out.reshape(-1, self.num_heads * self.v_head_dim)
+
     # NOTE(Xinyu): Make the loaded weight contiguous to avoid the transpose
     # during each graph execution
     def process_weights_after_loading(self, act_dtype: torch.dtype):
--- a/vllm_gaudi/models/deepseek_v2.py
+++ b/vllm_gaudi/models/deepseek_v2.py
@@ -3,6 +3,9 @@ from itertools import islice
 
 from vllm.distributed import get_pp_group
 from vllm.model_executor.models import deepseek_v2
+from vllm.model_executor.models.deepseek_v2 import (
+    DeepseekV32IndexerCache, Indexer)
+from vllm.model_executor.layers.sparse_attn_indexer import SparseAttnIndexer
 from vllm.sequence import IntermediateTensors
 
 
@@ -87,24 +90,65 @@ def _hpu_deepseek_v2_model_forward(
 # Applies to DeepseekV2/V3/Deepseek/GlmMoe/DSA — all share model_cls = DeepseekV2Model.
 deepseek_v2.DeepseekV2Model.forward = _hpu_deepseek_v2_model_forward
 
-_orig_deepseek_v2_model_load_weights = deepseek_v2.DeepseekV2Model.load_weights
+
+# ---------------------------------------------------------------------------
+# DSA / Indexer enablement on HPU
+# ---------------------------------------------------------------------------
+
+# --- IndexerCache: BF16 storage instead of FP8 uint8 -----------------------
+_orig_indexer_cache_init = DeepseekV32IndexerCache.__init__
 
 
-def _hpu_deepseek_v2_model_load_weights(self, weights):
-    """Drop GLM-5 DSA shared-indexer projection weights (`indexers_proj`).
-
-    vLLM's DeepseekV2Model has no module for them, and on HPU DSA layers run
-    as dense MLA with the indexer never executed, so the projection that
-    shares indexer K caches across layers is dead weight here.
-    """
-
-    def _filtered(ws):
-        for name, weight in ws:
-            if ".indexers_proj." in name:
-                continue
-            yield name, weight
-
-    return _orig_deepseek_v2_model_load_weights(self, _filtered(weights))
+def _hpu_indexer_cache_init(self, head_dim, dtype, prefix, cache_config):
+    if dtype == torch.uint8:
+        head_dim = head_dim * 128 // (128 + 4)
+        dtype = torch.bfloat16
+    _orig_indexer_cache_init(self, head_dim, dtype, prefix, cache_config)
 
 
-deepseek_v2.DeepseekV2Model.load_weights = _hpu_deepseek_v2_model_load_weights
+DeepseekV32IndexerCache.__init__ = _hpu_indexer_cache_init
+
+
+def _hpu_indexer_cache_get_attn_backend(self):
+    from vllm_gaudi.attention.backends.hpu_attn import HPUMLAAttentionBackend
+    return HPUMLAAttentionBackend
+
+
+DeepseekV32IndexerCache.get_attn_backend = _hpu_indexer_cache_get_attn_backend
+
+
+# --- Indexer.forward: BF16 path, skip FP8 quantization ---------------------
+def _hpu_indexer_forward(self, hidden_states, qr, positions, rotary_emb):
+    q, _ = self.wq_b(qr)
+    q = q.view(-1, self.n_head, self.head_dim)
+    q_pe, q_nope = torch.split(
+        q, [self.rope_dim, self.head_dim - self.rope_dim], dim=-1)
+    kw, _ = self.wk_weights_proj(hidden_states)
+    kw = kw.reshape(-1, kw.shape[-1])
+    k, weights = torch.split(kw, [self.head_dim, self.n_head], dim=-1)
+    k = self.k_norm(k.contiguous())
+    k_pe, k_nope = torch.split(
+        k, [self.rope_dim, self.head_dim - self.rope_dim], dim=-1)
+    q_pe, k_pe = rotary_emb(positions, q_pe, k_pe.unsqueeze(1))
+    q_pe = q_pe.reshape(-1, self.n_head, self.rope_dim)
+    k_pe = k_pe.reshape(-1, self.rope_dim)
+    k_nope = k_nope.reshape(-1, self.head_dim - self.rope_dim)
+    q = torch.cat([q_pe, q_nope], dim=-1)
+    k = torch.cat([k_pe, k_nope], dim=-1)
+    weights = weights.reshape(-1, self.n_head) * self.softmax_scale * self.n_head_scale
+    return self.indexer_op(hidden_states, q, k, weights)
+
+
+Indexer.forward = _hpu_indexer_forward
+
+
+# --- SparseAttnIndexer: dispatch to HPU forward -----------------------------
+_orig_forward_native = SparseAttnIndexer.forward_native
+
+
+def _hpu_sparse_indexer_forward_native(self, hidden_states, q_quant, k, weights):
+    from vllm_gaudi.ops.hpu_sparse_attn_indexer import forward_hpu
+    return forward_hpu(self, hidden_states, q_quant, k, weights)
+
+
+SparseAttnIndexer.forward_native = _hpu_sparse_indexer_forward_native
--- a/vllm_gaudi/ops/hpu_sparse_attn_indexer.py
+++ b/vllm_gaudi/ops/hpu_sparse_attn_indexer.py
@@ -1,0 +1,43 @@
+import torch
+from vllm.forward_context import get_forward_context
+
+
+@torch.compiler.disable
+def forward_hpu(self, hidden_states, q, k, weights):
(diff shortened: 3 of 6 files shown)


vllm_gaudi/attention/backends/hpu_attn.py
```diff
@@ -353,6 +357,28 @@ def forward_mqa(  # type: ignore
...
+        """
+        if isinstance(k_cache, tuple):
+            k_cache = k_cache[0]
+        batch_size = q.shape[0]
+        topk = topk_indices.shape[1]
+        topk_indices = topk_indices[:batch_size]
+        pad_mask = topk_indices == -1
+
+        flat_idx = topk_indices.clamp(min=0).reshape(-1)
+        selected = k_cache[flat_idx].view(batch_size, topk, self.kv_lora_rank + self.qk_rope_head_dim)
```
pawel-olejniczak, line 380: `forward_mqa_sparse` reads the latent cache directly, which makes it the only MLA path that does not go through `self.latent_cache_k.fetch_from_cache`. `forward_mha` uses it at :277 and `forward_mqa` hands it over as `keys_fetch_func` at :355. That matters under `--kv-cache-dtype fp8_inc`, because the fetch wrapper is where dequantization lives:

* with `QUANT_CONFIG` unset, `enable_fp8_attn` is True (:214) and `latent_cache_k` is a `VLLMFP8KVCache` (:224), whose `fetch_from_cache` ends in `dequant_output` (`vllm_gaudi/extension/utils.py:114-121`);
* with `QUANT_CONFIG` set it stays a plain `VLLMKVCache`, but INC's `PatchedVLLMKVCache` wraps that same method, per the note at `vllm_gaudi/extension/utils.py:84-85`.

The write side still quantizes either way, since `oot_mla.py:163` goes through `self.impl.latent_cache_k(...)`, and `hpu_model_runner.py:6893-6902` allocates the tensor with `dtype = kv_cache_spec.dtype`, which is `float8_e4m3fn` for `fp8_inc`. So `k_cache[flat_idx]` on this line yields raw fp8 rows and the `torch.bmm` two lines down pairs them with a bf16 query.

I grepped `platform.py`, `oot_mla.py` and `extension/features.py` for a guard on `use_sparse` together with `fp8_inc` and found none, so I believe `--kv-cache-dtype fp8_inc` on a DSA model reaches this line and faults on the dtype mismatch at the first decode step.

One caution if you decide to fix rather than reject: routing this through `fetch_from_cache` is not a drop-in. Under contiguous PA that method takes the `cache.narrow(0, 0, blocks.size(0))` identity fast path (`vllm_gaudi/extension/utils.py:86-91`), which is only valid for the block-major decode layout, not for the arbitrary physical slot indices used here. An explicit dequant, or a `NotImplementedError` in `HPUPlatform.get_attn_backend_cls` alongside the existing `use_sparse and not use_mla` check, both look safer.

Is fp8 KV cache meant to be in scope for DSA at this stage? If not, an explicit guard would at least keep the failure legible.

vllm_gaudi/ops/hpu_sparse_attn_indexer.py
```diff
@@ -0,0 +1,71 @@
...
+    valid = valid & ((block_groups >= 0) & (block_groups < batch_size))[:, None]
+    keys = kv_cache[slots.reshape(-1)].reshape(block_count, block_size, -1).float()
+    owners = block_groups.clamp(0, batch_size - 1)
+    block_query = q[owners].float()
+    block_weights = weights[owners].float()
+    logits = torch.bmm(block_query, keys.transpose(1, 2))
+    scores = (logits.relu() * block_weights[:, :, None]).sum(1)
+    rows = torch.arange(batch_size, device=block_list.device)
+    mask = (block_groups[None, :, None] == rows[:, None, None]) & valid[None, :, :]
+    scores = scores[None, :, :].expand(batch_size, -1, -1).masked_fill(~mask, -torch.inf)
```
pawel-olejniczak, line 61: This line materializes a `[batch_size, block_count, block_size]` fp32 tensor, and :61 materializes a bool mask of the same shape, even though each block belongs to exactly one request, so at most `1 / batch_size` of it is ever finite.

Taking `block_size = 128` and an indexer with `head_dim = 128` and `n_head = 64`, at `max_num_seqs = 32` with 32k context per request (`block_count = 8192`), the fp32 transients live at this point are:

* `keys` (:54): `8192 * 128 * 128 * 4` = 512 MiB
* `block_query` (:56): `8192 * 64 * 128 * 4` = 256 MiB
* `logits` (:58): `8192 * 64 * 128 * 4` = 256 MiB
* `mask` (:61): `32 * 8192 * 128` = 32 MiB
* this line: `32 * 8192 * 128 * 4` = 128 MiB

About 1.2 GiB alive at once, per DSA layer, per decode step, followed by a `topk` over roughly 1M candidates per row. The two batch-scaled terms are what the vectorized rewrite added.

I do not think the 128k needle runs you described on #1777 would have surfaced this. At batch 1 with `block_count = 1024` the same five tensors come to about 129 MiB, so the batch-scaled terms are two orders of magnitude smaller there.

I looked for a cheaper formulation and did not find a clean one: a per-request top-k over blocks is a segmented top-k, and `block2batch` (`vllm_gaudi/extension/ops.py:90`) is a sum reduction, so the `flat_pa` trick does not carry over. The one lever I do see is the `.float()` on :54, :56 and :57, which doubles the three largest tensors. Note that fp32 softmax is a per-model feature flag in this repo (`vllm_gaudi/extension/features.py:88`) rather than an unconditional choice.

Have you measured headroom with high concurrency and long context together, for example `max_num_seqs` 32 at 32k? If fp32 scoring is needed for accuracy that is fine; I would mainly like to know where the ceiling is, so it can be documented or bucketed against rather than discovered as an OOM.

pawel-olejniczak (commented):
Reviewed at `0052c82` against `VLLM_STABLE_COMMIT` `d2906091`, resolved through the `vllm/last-good-commit-for-vllm-gaudi` branch per `.github/workflows/pre-merge.yaml:275-283`. This is a follow-up to the review on the closed backport #1777, so I have skipped what was settled there: the three gates that kept the sparse path unreachable are gone, and the validation and performance questions were answered.

The vectorized rewrite in `0052c82` reads well and I could not break the slot arithmetic. The indexer cache, the MLA latent cache and `_PAD_BLOCK_ID` all agree on the flat `(num_blocks + 1) * block_size` numbering that `HPUMLAAttentionBackend.get_kv_cache_shape` establishes at `hpu_attn.py:113-120`, so the `_PAD_BLOCK_ID` rows that `slots` pulls in are in bounds; the `block_groups == -1` and `block_usage == 1` padding is masked out correctly by the mask on `hpu_sparse_attn_indexer.py:61`; the `forward_native` patch target is right, since the out-of-tree `CustomOp` dispatch lands on the base `forward_oot`, which calls it; and `_hpu_indexer_forward` matches upstream's generic branch line for line minus the q FP8 quant. The new `AttentionLayerBase` branch in the runner is also narrower than what upstream does at `gpu_model_runner.py:7644-7663`.

Four things I would like resolved, in descending order of how much they worry me, plus one mechanical comment.

vllm_gaudi/extension/features.py
```diff
@@ -97,7 +97,9 @@ def get_features():
         ValueFromList('prompt_attn_impl', supported_attn_impls),
         Value('skip_warmup', False),
         Value('merged_prefill', False),
-        Value('use_contiguous_pa', Disabled('prefix_caching'), env_var='VLLM_CONTIGUOUS_PA'),
+        Value('use_contiguous_pa',
+              All(Disabled('prefix_caching'), Not(ModelType('glm_moe_dsa'))),
```
pawel-olejniczak: Two gaps in this guard, both of which leave the indexer scoring the full allocated block range rather than the blocks in use.

1. The env var outranks the computed default. `Value.__call__` returns the override before it ever evaluates the dependency expression (`vllm_gaudi/extension/config.py:196-204`), so `VLLM_CONTIGUOUS_PA=true` re-enables contiguous PA for exactly the model this clause excludes. Under contiguous PA, `block_bucket_size` is derived from `max(block_list) + 1` rather than from the number of blocks the batch actually holds (`hpu_model_runner.py:2457-2464`), so `block_count` in the indexer tracks the high water mark of the whole cache instead of the live batch.

2. The guard is one exact model_type. `ModelType` is `Eq('model_type', ...)` (`vllm_gaudi/extension/config.py:110-112`) against `hf_config.model_type` (`vllm_gaudi/extension/environment.py:75`), so it covers `glm_moe_dsa` only. DeepSeek-V3.2 reaches the same new sparse path: `is_sparse` is `is_v32`, which upstream sets from `hasattr(config, "index_topk")` (`vllm/model_executor/models/deepseek_v2.py:1421` and `:1200`), and on a non-CUDA platform `DeepseekV32ForCausalLM` resolves to `DeepseekV3ForCausalLM` from `deepseek_v2` (`vllm/models/deepseek_v32/__init__.py:26-29`), so it goes through the same new `oot_mla.py:168` branch with `model_type == "deepseek_v32"` and keeps contiguous PA on.

The commit message for `7d2dfa2` does not say whether disabling contiguous PA here was a correctness fix or a memory one, and that changes the answer. If it is memory, would `Not(Any(ModelType('glm_moe_dsa'), ModelType('deepseek_v32')))` be the right shape here? If it is correctness, a feature default that a single env var can defeat seems like the wrong mechanism, and a check that raises would hold better.

vllm_gaudi/models/deepseek_v2.py
```diff
@@ -87,24 +89,76 @@ def _hpu_deepseek_v2_model_forward(
 # Applies to DeepseekV2/V3/Deepseek/GlmMoe/DSA — all share model_cls = DeepseekV2Model.
 deepseek_v2.DeepseekV2Model.forward = _hpu_deepseek_v2_model_forward
 
-_orig_deepseek_v2_model_load_weights = deepseek_v2.DeepseekV2Model.load_weights
+# ---------------------------------------------------------------------------
+# DSA / Indexer enablement on HPU
+# ---------------------------------------------------------------------------
 
+# --- IndexerCache: BF16 storage instead of FP8 uint8 -----------------------
+_orig_indexer_cache_init = DeepseekV32IndexerCache.__init__
```
pawel-olejniczak, line 97: This hunk removes `_hpu_deepseek_v2_model_load_weights`, whose only job was to drop `.indexers_proj.` weights before upstream `load_weights` saw them. Its docstring gave two reasons: that DSA layers ran as dense MLA here, and that "vLLM's DeepseekV2Model has no module for them". This PR invalidates the first reason. I cannot find anything that invalidates the second.

At `d2906091` there is no `indexers_proj` anywhere in the vLLM tree, and I do not see a path through `DeepseekV2Model.load_weights` that would skip such a name:

* the `.indexer.` skip at `vllm/model_executor/models/deepseek_v2.py:1630-1633` tests for the literal `".indexer."`, which `".indexers_proj."` does not contain;
* `maybe_remap_kv_scale_name` returns the name untouched unless it ends in a scale or zero-point suffix (`vllm/model_executor/model_loader/weight_utils.py:1464-1495`);
* so control reaches `param = params_dict[name]` at `vllm/model_executor/models/deepseek_v2.py:1793` and raises `KeyError`.

On HPU this is the live code path, since `GlmMoeDsaForCausalLM` resolves to the `deepseek_v2` class on a non-CUDA platform (`vllm/models/deepseek_v32/__init__.py:29`), which is presumably why the filter was added in #1717 in the first place.

Your accuracy runs clearly load the model, so something in my reading is wrong. What handles those weights now? If the GLM-5.2-FP8 revision you tested simply does not ship `indexers_proj`, that is worth a line in the PR description, because the filter was added because some revision did.

vllm_gaudi/attention/backends/hpu_attn.py
```diff
@@ -260,6 +260,10 @@ def __init__(
                                                  f"heads in the layer. Sinks shape: {sinks.shape}, "
                                                  f"num_heads: {num_heads}.")
 
+
```
pawel-olejniczak: Grouping the mechanical items here so they do not clutter the rest.

1. Two gating checks are red or will go red at `0052c82`.

   `pre-commit` fails, and `pre_merge_hpu_test_build` has `needs: [pre-commit, ...]` (`.github/workflows/pre-merge.yaml:238-240`), so the HPU suite will not run until it is green. The pinned yapf 0.43.0 rewrites exactly two hunks; pinned ruff 0.11.7 is clean. The first is the extra blank line on this line, the second is `oot_mla.py:170-171`, which yapf joins into one 112-column line:

   ```
   output = self.impl.forward_mqa_sparse(query, kv_cache, attn_metadata, self.impl.topk_indices_buffer)
   ```

   `DCO` is `action_required` on three commits: `5c99ade` signs off under a different address than its author field, and `23f76af` and `52ea3f7` have no sign-off at all. The check output on the PR lists the exact expected values.

2. `_fill_invalid(buf, n, device)` in `hpu_sparse_attn_indexer.py:7` never uses `device`, and all three call sites pass `q.device`. Worth dropping the parameter.

3. `hpu_sparse_attn_indexer.py:53` narrows `valid` by `(block_groups >= 0) & (block_groups < batch_size)`, but `valid` is only consumed by the mask on :61, where `block_groups[None, :, None] == rows[:, None, None]` already excludes every block group outside `[0, batch_size)`. The clause looks redundant.

4. `platform.py:90` expands DSA as "Dynamic Sparse Attention". Upstream calls it DeepSeek Sparse Attention (`vllm/models/deepseek_v32/__init__.py:5-7`), and the new comment on `platform.py:103` in this PR uses the acronym without expanding it.

vllm_gaudi/models/deepseek_v2.py
jkaniecki: As I mentioned in 0.28.0 PR:

I'm not sure what was the purpose of this filter before - I've checked the GLM-5.2-FP8 model and it doesn't contain any indexers_proj

grep -o 'indexers_proj' model.safetensors.index.json | wc -l
→ 0
It contains indexer projestion weight under:
model.layers.N.self_attn.indexer.weights_proj.weight

The previous MLA fallback was also tested with GLM5.2-FP8 model, so not sure what was the goal of this filter (maybe some previous versions of FP8 model have included such weights). When using newest version of this model there are no longer `indexers_proj` weights, so this filter is not needed. Added a not in PR message

vllm_gaudi/attention/backends/hpu_attn.py
jkaniecki: fp8 kv cache is not a part of this enablement and has not been requested for this model - maybe future optimizations will lead to fp8 kv cache feature. Added a proper guard in `HPUMLAImpl. __init__`

vllm_gaudi/extension/features.py
jkaniecki: This was made for corectness. We don't want the defrag (which is on by default) to mess with indexer cache. Most of the time we use prefix caching with GLM, so cont PA is off anyway. But for cases when user turns prefix caching off I will add a guard that will inform about cont pa being unsupported with GLM. Maybe in future adjustments we will add support or remove defrag from DSA and indexer cache, but for now just let's reject those mechanisms for GLM. Added a guard for cont PA and defrag.

vllm_gaudi/ops/hpu_sparse_attn_indexer.py
jkaniecki: Due to memory concerns (kv cache + very large model) we practically won't fit bs > 12/16 on single HLS (with ctx 32k). Most users want higher contexts to be functional, which means lowering the batch size further. I think that our memory will not suite more than 3500-4000 blocks, so I wouldn't focus on tensors memory here as much. Moreover, currently we are struggling with some long context (>128k) acc issues - till it's resolved I would leave fp32 calculations here to make sure it's numerically stable.

pawel-olejniczak (approved):
LGTM. Please check code formatting

iboiko-habana (approved):

PR #1761, opened 2026-08-27
[HPU] Make sampling penalties correct under async scheduling

jiminha:
Repetition/presence/frequency penalties read each request's output_token_ids, but under async scheduling the real sampled id is still being copied to CPU when the next step's sampler runs, so the list is stale. Two problems compound on HPU:

1. Correctness. Append a -1 placeholder to output_token_ids after sampling, then splice the real id over it right before the sampler reads the list for penalties. Ported from vLLM's GPU runner (InputBatch.set_async_sampled_token_ids / update_async_output_token_ids in vllm/v1/worker/gpu_input_batch.py, plus the placeholder append in gpu_model_runner.py).

2. Throughput. Under torch.compile, tracing apply_logits_processors installs a dynamo guard on len(output_token_ids[0]); that length grows every decode step and forces a full sampler recompile per step once penalties are active. Graph-break apply_logits_processors so the penalty math runs eagerly while sample() stays compiled.

Both the placeholder bookkeeping and the per-step CPU sync are skipped when no penalties are active, so the default fast path is unchanged.

Found while running OmniDocBench: on some dense, table-heavy pages the model looped and never emitted EOS. A repetition penalty fixed it on CUDA but made no difference at all on HPU -- because under async scheduling the penalty was reading an output_token_ids list that was never populated.


jiminha:
Relevant upstream PR is 
https://github.com/vllm-project/vllm/pull/26467/
https://github.com/vllm-project/vllm/pull/27910/

vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -4670,6 +4672,16 @@ def sample_tokens(self, grammar_output: "GrammarOutput | None") -> ModelRunnerOu
                 req_id: i
                 for i, req_id in enumerate(self.input_batch.req_ids) if i not in invalid_req_indices_set
             }
+            # Async scheduling: append a -1 placeholder to each request's
+            # output_token_ids so penalties see the right length (real id still
+            # copying to CPU; spliced in next step by
+            # update_async_output_token_ids). Skipped when penalties are off to
+            # keep the fast path free of the per-step sync. Mirrors the append
+            # in vllm/v1/worker/gpu_model_runner.py.
+            if not self.input_batch.no_penalties:
```
pawel-olejniczak: The placeholder append is gated on `no_penalties`, but upstream gates on `needs_output_token_ids` (penalties **or** bad_words **or** logitsprocs, `gpu_input_batch.py:910`) and appends the placeholder *unconditionally* - only the repair is conditional (`gpu_model_runner.py:3890-3917`).

That difference matters more here than on GPU. As far as I can tell this mechanism is the only thing that populates `output_token_ids` under async scheduling on HPU: `_update_states` only extends it on `not is_last_rank` (1901-1913), and the other append at 4754 iterates `postprocessed_sampled_token_ids`, which stays a list of empty lists in the async branch (4664-4685). So it isn't that penalties were one token behind - they were reading an empty list.

Since `no_penalties` is a batch-wide property (`hpu_input_batch.py:859`), the consequence is that whether a request's `output_token_ids` gets maintained at all depends on whether some *other* request in the batch happens to use penalties. Anything else reading that list (`bad_words` today, via `sampler.py:398`) then works or doesn't depending on unrelated traffic, which is a nasty thing to debug.

I'd just append unconditionally in the async branch and keep the gate on the CPU sync + splice, like upstream. Same fast path, no cross-request coupling.

tests/full_tests/test_async_penalty_consistency.py
```diff
@@ -0,0 +1,123 @@
...
+        del llm
+        gc.collect()
+
+
+@pytest.fixture(scope="module")
+def async_vs_sync():
+    """Ground-truth (async off) and async-on outputs for every config."""
+    reference = _run_all(async_scheduling=False)
+    async_out = _run_all(async_scheduling=True)
+    return reference, async_out
```
pawel-olejniczak, line 104: `_run_all` computes the no-penalty baseline for both engines, but nothing ever compares `reference["none"]` against `async_out["none"]`. That comparison is the control for the docstring's central claim ("any mismatch is unambiguously the penalty staleness") - async scheduling shifts step boundaries and batch composition, so it's worth proving async and sync are token-identical *without* penalties before attributing a diff to penalties. The data is already there; one extra assert and the test stops being able to misattribute a bucketing/padding difference.

vllm_gaudi/utils.py
```diff
@@ -233,6 +233,18 @@ def make_tensor_with_pad_hpu(
...
+    # apply_logits_processors reads the per-request output_token_ids lists,
+    # whose lengths grow by one each decode step; tracing that access installs
+    # a dynamo guard on len(output_token_ids[0]) that forces a full sampler
+    # recompile every step once penalties are active. Disabling compile here
+    # runs the penalty/bad-words math eagerly (no length guard) while sample()
+    # stays compiled. No-op on backends that run the sampler eagerly (e.g. CUDA).
+    from vllm.v1.sample.sampler import Sampler
+
+    Sampler.apply_logits_processors = torch.compiler.disable(  # type: ignore[method-assign]
+        Sampler.apply_logits_processors)
```
pawel-olejniczak: Two things on the graph break:

Regional compilation is on by default and that's the only path that compiles the sampler (`hpu_model_runner.py:4986-4990`), so with `VLLM_T_COMPILE_FULLGRAPH=true` this becomes a disabled callee inside a `fullgraph=True` compile - which should be an error rather than a break. Did you try that combination? If it does blow up, gating the patch on `fullgraph` being off (or compiling the sampler with `fullgraph=False`) would keep it contained.

Also, since throughput is the whole justification for globally patching a vLLM class, could you add a before/after number (step time or tok/s, with and without penalties, in compile mode)? The test runs `enforce_eager=True`, so this half of the PR is currently unmeasured and untested.

Minor: the patch sits in the `not is_lazy()` block next to the `make_tensor_with_pad` override, so it also applies when nothing compiles the sampler. Next to `self.sampler = self._compile(self.sampler)` would tie it to the condition it actually depends on.

tests/full_tests/test_async_penalty_consistency.py
```diff
@@ -0,0 +1,123 @@
+# SPDX-License-Identifier: Apache-2.0
```
pawel-olejniczak, line 1: This will run twice per PR: the e2e matrix is generated by grepping `^run_` out of `ci_e2e_discoverable_tests.sh` (`pre-merge.yaml:171`), so `run_async_penalty_consistency_test` already gets its own job, and `hpu_unit_tests` pytests the whole `tests/unit_tests` tree (`pre-merge.yaml:381`). For a test that builds two engines, `tests/full_tests/` looks like the better home - and it keeps the two engines out of the shared unit-test process.

Also worth running `yapf` before CI gets approved - pinned 0.43.0 wants two changes in this file (the `enforce_eager=True` inline comment and the assert continuation in `test_penalties_are_active`). The other two files are clean.

pawel-olejniczak (commented):
Went through this with the async-scheduling path in mind. The core idea is right and the port of `update_async_output_token_ids` is faithful, including the placeholder-count reconciliation. Four things worth a look before merge, one of them substantive.

vllm_gaudi/v1/worker/hpu_model_runner.py
jiminha: Updated.

tests/full_tests/test_async_penalty_consistency.py
jiminha: Updated

vllm_gaudi/utils.py
jiminha: You are right, VLLM_T_COMPILE_FULLGRAPH does system crash. I will compile the sampler with fullgraph=False. 
Without this fix, the decode time is unbearably slow. I will measure it with Gemma4 use case and will update.

tests/full_tests/test_async_penalty_consistency.py
jiminha: Updated

vllm_gaudi/utils.py
jiminha: I tested with Gemma-4 E4B, TP1, torch-compile mode (PT_HPU_LAZY_MODE=0), async scheduling on.
vllm bench serve, 200 prompts, output length 380, max concurrency 45.
Without graph break with penalty on, it's extremely slow. 

| tree | no penalty | penalty (rep 1.1) |
|---|---|---|
| baseline `7067d98^` | 461.31 tok/s | 416–417 tok/s (penalty inert, cost paid anyway) |
| fix + graph break  | 462.29 tok/s | 390.77 tok/s |
| fix + no graph break | 462.83 tok/s | 5.30 tok/s (recompile storm) |

pawel-olejniczak (approved):

iboiko-habana (approved):

PR #1763, opened 2026-08-28
Migrate model weight-loader skip rules, fix for #53106

iboiko-habana:
## Summary

Update Gaudi model weight loading for the upstream `AutoWeightsLoader` API change
that removes `skip_prefixes` and `skip_substrs`.

- Replace loader skip options in unit tests with `WeightsMapper` mappings to
  `None`, preserving coverage for filtering unwanted checkpoint weights.
- Remove obsolete tied-`lm_head` skip handling from GPTBigCode and StarCoder2.
  The upstream loader now handles aliased/tied parameters directly.

## Upstream Compatibility

Required after upstream PR #53106 removes `AutoWeightsLoader` constructor
arguments `skip_prefixes` and `skip_substrs`.


pawel-olejniczak (commented):
Reviewed this against a vLLM checkout that contains vllm-project/vllm#53106, and verified the behaviour empirically rather than only by reading the diff. The fix is correct and sufficient for the reported failure -- two remarks inline.

What I confirmed as fine:

- `AutoWeightsLoader.__init__` really is down to `ignore_unexpected_prefixes` / `ignore_unexpected_suffixes`, so the change is required.
- The migration is complete: no other `skip_prefixes` / `skip_substrs` remain in the repo, and the other vendored models touched by the same upstream commit are unaffected (`minimax_m2`'s MTP skip is hand-rolled in `get_spec_layer_idx_from_weight_name`, not via the removed kwargs).
- Both rewritten tests pass.
- Tied-`lm_head` semantics with the idiom these files use (`self.lm_head = self.transformer.wte`) are equivalent to the old skip: aliases resolve to `{'lm_head.weight': 'transformer.wte.weight'}`, only the embedding name ends up in `loaded`, `lm_head` takes the embedding's values, and the result is independent of weight order in the checkpoint stream. The untied path is unchanged.
- The reported checkpoint has no `lm_head.*` keys at all, so the crash was purely the `TypeError` at loader construction (which fires regardless of `tie_word_embeddings`, since the kwarg was always passed), and the load is clean afterwards.
- `VLLM_STABLE_COMMIT` already contains the culprit, so pre-merge does build against the new API and no `[FIX_FOR_VLLM_CUSTOM=...]` tag is needed.

One more note for the bug report rather than the code: a tied checkpoint that ships only `lm_head.weight` (no embedding key) now raises `ValueError: 'lm_head.weight' was skipped because it is tied to '<embed>' ... so the tied weight is uninitialized.` where it previously left the embedding silently uninitialized. That is strictly better, but it is a new message worth recording so it is not triaged as a fresh regression later.

vllm_gaudi/models/gpt_bigcode.py
```diff
@@ -323,11 +323,5 @@ def compute_logits(
...
 
     def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
-        skip_prefixes = None
-        if self.config.tie_word_embeddings:
-            skip_prefixes = ["lm_head."]
-        loader = AutoWeightsLoader(
-            self,
-            skip_prefixes=skip_prefixes,
-        )
+        loader = AutoWeightsLoader(self)
```
pawel-olejniczak: **Behaviour change worth guarding: the alias handling is narrower than the old prefix skip.**

The upstream tied-parameter handling drops **only the aliased parameter itself** (`lm_head.weight`). The old `skip_prefixes=["lm_head."]` matched with `startswith`, so it also dropped every *other* `lm_head.*` key in the checkpoint.

I verified this against a vLLM that contains vllm-project/vllm#53106, using a module tied the same way this file ties it (`self.lm_head = self.transformer.wte`):

```
lm_head.weight        -> skipped, embedding wins           (same as before)
lm_head.bias          -> tolerated (automatic ".bias" in ignore_unexpected_suffixes)
lm_head.weight_scale  -> ValueError: There is no module or parameter named
                         'lm_head.weight_scale' in ... The available parameters
                         belonging to lm_head (VocabParallelEmbedding) are:
                         {'lm_head.weight'}
```

So a **quantized** GPTBigCode checkpoint with tied embeddings that ships `lm_head.*` scale tensors goes from "loads fine" to a hard load failure. This file does pass `quant_config` down and already has an fp8 `c_attn.input_scale` path, so it is not a purely hypothetical shape.

The bf16 `granite-20b-code-instruct-8k` checkpoint from the bug report is unaffected: it has **no `lm_head.*` keys at all** (628 weights, `transformer.wte.weight` present), so the old skip was a no-op there and this PR fixes it correctly. Not a blocker, but the guard is one line and keeps the old tolerance without reintroducing a removed kwarg:

```suggestion
        loader = AutoWeightsLoader(
            self,
            # A tied checkpoint may still carry other lm_head.* tensors (e.g. quant
            # scales); only the tied weight is dropped by the loader's alias handling.
            ignore_unexpected_prefixes=(["lm_head."] if self.config.tie_word_embeddings else None),
        )
```

Keeping it conditional on `tie_word_embeddings` matters -- unconditionally ignoring `lm_head.` would mask a genuinely missing `lm_head` in the untied path.

vllm_gaudi/models/starcoder2.py
```diff
@@ -349,10 +349,5 @@ def compute_logits(
         return logits
 
     def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
-        loader = AutoWeightsLoader(
-            self,
-            # Models trained using ColossalAI may include these tensors in
-            # the checkpoint. Skip them.
-            skip_prefixes=(["lm_head.weight"] if self.config.tie_word_embeddings else None),
-        )
+        loader = AutoWeightsLoader(self)
```
pawel-olejniczak: Same point as in `gpt_bigcode.py`. Here the old value was `skip_prefixes=["lm_head.weight"]`, which -- being a `startswith` match -- also covered `lm_head.weight_scale`, so the narrowing applies identically.

This also gives the deleted ColossalAI note somewhere to live, since it explained *why* the tolerance existed:

```suggestion
        loader = AutoWeightsLoader(
            self,
            # Models trained using ColossalAI may include extra lm_head.* tensors in
            # the checkpoint; only the tied weight is dropped by the alias handling.
            ignore_unexpected_prefixes=(["lm_head."] if self.config.tie_word_embeddings else None),
        )
```

tests/unit_tests/test_auto_weights_loader.py
```diff
@@ -97,8 +97,11 @@ def weight_generator():
     assert not torch.all(new_mod.nested_mod.bn.running_var == mod.nested_mod.bn.running_var)
     assert new_mod.nested_mod.bn.num_batches_tracked.item() == 0
 
-    loader = AutoWeightsLoader(new_mod, skip_prefixes=["prefix."])
-    loader.load_weights(weight_generator())
+    loader = AutoWeightsLoader(new_mod)
+    loader.load_weights(
+        weight_generator(),
+        mapper=WeightsMapper(orig_to_new_prefix={"prefix.": None}),
+    )
```
pawel-olejniczak, line 104: **Three things converge here, and I think the test effort is better spent elsewhere.**

**1. This file is not collected by any CI job.** Every pytest entry point in `.github/workflows/{pre-merge,hourly-ci,create-release-branch}.yaml` targets `tests/unit_tests`; the e2e scripts target specific files under `tests/models/language|multimodal`. `tests/models/test_utils.py` is not referenced anywhere in the repo. So this migration is currently dead code -- it will never run, pass or fail.

**2. Upstream deleted these two tests in the very same commit.** vllm-project/vllm#53106 removed `test_module_skip_prefix` and `test_module_skip_substr` (-76 lines) and replaced them with `test_module_skip_tied_weights` and `test_get_rename_mapper_keeps_only_renames`. Keeping them here means vllm-gaudi starts maintaining a fork-local test of pure upstream `WeightsMapper` behaviour, with no HPU-specific content -- permanent sync debt for zero signal. (The names and docstrings also still describe an API that no longer exists; the substr test's docstring has always read *"can skip prefix"*.)

**3. Nothing covers what this PR actually changes.** The production change is the tied-`lm_head` removal in `gpt_bigcode.py` / `starcoder2.py`, and no test touches it. Upstream's replacement `test_module_skip_tied_weights` will not help us either: it ties via `ParallelLMHead.tie_weights()`, whereas both of our files tie by **module aliasing** (`self.lm_head = self.transformer.wte`). Different structure -- so this idiom is covered nowhere, upstream or here.

That gap is not academic. `VLLM_STABLE_COMMIT` (`0ecc2847`, 2026-08-24) already contained the culprit (`1fe3a15`, 2026-08-20) for four days and pre-merge stayed green -- the break only surfaced in downstream QA validation. And per the comment at `vllm_gaudi/models/__init__.py:66-71`, upstream now routes both architectures to the Transformers backend, so these two files are **permanently orphaned forks** with no upstream sibling to sync from. Gaudi-local coverage is the only guard we will ever have here.

Suggestion: drop both migrated tests (mirroring upstream) and add the guard below under `tests/unit_tests/`, where CI actually runs pytest and where the `dist_init` / `default_vllm_config` fixtures already exist in `conftest.py`. I ran it against a vLLM containing #53106 -- **2 passed** -- and confirmed it is not vacuous: stubbing out `_get_tied_embedding_params` makes the `tie=True` case load both names and leaves `lm_head` at `2.0`, failing the assertions.

`tests/unit_tests/test_tied_lm_head_weight_loading.py`:

```python
# SPDX-License-Identifier: Apache-2.0
# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
"""Weight-loading guard for the vendored GPTBigCode / Starcoder2 models.

Both tie `lm_head` to the input embedding by *module aliasing*
(`self.lm_head = self.transformer.wte`) instead of
`ParallelLMHead.tie_weights()`, so upstream's own tied-weight test does not
cover this shape. Upstream has dropped its copies of these architectures, so
this is the only regression guard we get.
"""

import pytest
import torch

from vllm.model_executor.layers.vocab_parallel_embedding import (ParallelLMHead, VocabParallelEmbedding)
from vllm.model_executor.models.utils import AutoWeightsLoader

VOCAB_SIZE = 16
HIDDEN_SIZE = 2


class AliasTiedLMHead(torch.nn.Module):
    """Mimics how GPTBigCodeForCausalLM / Starcoder2ForCausalLM tie lm_head."""

    def __init__(self, tie: bool):
        super().__init__()
        self.transformer = torch.nn.Module()
        self.transformer.wte = VocabParallelEmbedding(VOCAB_SIZE, HIDDEN_SIZE)
        self.lm_head = (self.transformer.wte if tie else ParallelLMHead(VOCAB_SIZE, HIDDEN_SIZE))


def _embedding(value: float) -> torch.Tensor:
    return torch.full((VOCAB_SIZE, HIDDEN_SIZE), value)


@pytest.mark.parametrize("tie", [True, False])
def test_tied_lm_head_is_loaded_once(default_vllm_config, dist_init, tie):
    """A tied `lm_head.weight` is dropped in favour of the input embedding."""
    model = AliasTiedLMHead(tie)
    weights = [
        ("transformer.wte.weight", _embedding(1.0)),
        ("lm_head.weight", _embedding(2.0)),
    ]

    loaded = AutoWeightsLoader(model).load_weights(iter(weights))

    if tie:
        assert loaded == {"transformer.wte.weight"}
        assert torch.all(model.lm_head.weight[:VOCAB_SIZE] == 1.0)
    else:
        assert loaded == {"transformer.wte.weight", "lm_head.weight"}
        assert torch.all(model.lm_head.weight[:VOCAB_SIZE] == 2.0)
```

If you take the `ignore_unexpected_prefixes` guard from the other comment, adding `("lm_head.weight_scale", torch.tensor([1.0]))` to the `tie=True` weight list turns this into a regression test for that too.

pawel-olejniczak, line 104: **The new filename promises coverage that isn't there.**

Commit 3 is a `similarity index 100%` rename — the contents are byte-identical to before. `tests/unit_tests/test_tied_lm_head_weight_loading.py` now holds four batchnorm tests (`test_module_with_batchnorm_can_load`, `test_module_with_child_containing_batchnorm_can_autoload`, `test_module_skip_prefix`, `test_module_skip_substr`) and not a single tied-`lm_head` test.

To be clear about what the move *did* fix: these tests are now collected by CI, which was the first of my three points, and renaming was genuinely necessary because `tests/unit_tests/test_utils.py` already exists. Thanks for that.

But on substance nothing changed. The mechanism this fix depends on — dropping the tied `lm_head.weight` alias when `lm_head` is tied by **module aliasing** (`self.lm_head = self.transformer.wte`) rather than by `ParallelLMHead.tie_weights()` — still has no test. Neither does the `ignore_unexpected_prefixes` branch added in commit 2.

I'd argue this is now worse than before the rename. The filename advertises a guard that does not exist, so the next person looking for tied-`lm_head` coverage finds a file with exactly the right name and concludes the case is handled. That is not a naming nit — it is a hidden gap, and these two models have no upstream sibling and no other coverage anywhere in CI, so a green pre-merge here carries no signal about them.

Two reasonable ways out:

1. Drop the test from my earlier review comment into this file. It is verified passing (2 passed) and parametrizes over tied/untied, so it also exercises the new `ignore_unexpected_prefixes` branch.
2. Or move the four batchnorm tests under a neutral name such as `test_auto_weights_loader.py`, and don't imply they cover tied `lm_head`.

Either works. What I'd like to avoid is a filename that claims the guard exists while the loader path this PR actually relies on stays untested.

pawel-olejniczak (approved):

PR #1766, opened 2026-08-28
Port of: [Bugfix][HPU] Fix Kimi-K2.5 vision 2D-RoPE warmup crash after upstream fused-kernel inlining (#50400) #1754

iboiko-habana:


shepark:
@iboiko-habana we don't need this changes in releases/v0.28.0 because there's no corresponding upstream vllm changes, which is https://github.com/vllm-project/vllm/commit/c8de51991 in vllm releases/v0.28.0.

iboiko-habana:
@shepark thanks

PR #1770, opened 2026-08-28
 Disable Synapse persistent-input reuse for compact-GDN

slokesha:
With compact-GDN (VLLM_COMPACT_GDN), the GDN recurrent-state (conv/ssm) read and write are split across separate torch.compile recipes. Synapse's per-graph persistent-input reuse can then reuse a still-live state buffer's memory as intra-graph scratch — in the recipe that only reads the state it looks like a dead intermediate, but another recipe / the next decode step still needs it — corrupting the state and producing NaN output (gsm8k 0.00, all-! tokens) on Qwen3.5-A17B FP8 at TP=8.

The reuse is opt-in per input and decided one recipe at a time, so neither Synapse nor the bridge's per-graph analysis can detect this cross-recipe reuse. This disables it on the compact-GDN path via the existing PT_HPU_ENABLE_SYNAPSE_INPUT_REUSE flag, set in set_compile_env_defaults() (torch.compile-only, no-op in lazy, and it never overwrites a user-set value). It only forgoes an input-DRAM-reuse memory optimization — no compute/latency impact.

Clone of https://github.com/vllm-project/vllm-gaudi/pull/1753 for v0.28.0


iboiko-habana:
New env vars adding procedure is next.
Please re-write it via gaudi_envs usage and add description to docs/configuration/env_variables.md
example: https://github.com/vllm-project/vllm-gaudi/pull/1744
https://github.com/vllm-project/vllm-gaudi/pull/1744#pullrequestreview-4981497176

vllm_gaudi/ops/hpu_gdn_pytorch.py
```diff
@@ -36,9 +36,18 @@
 _USE_EXACT_SOLVE = os.getenv("VLLM_GDN_EXACT_SOLVE", "0") == "1"
```
iboiko-habana, line 36: Please handle VLLM_GDN_EXACT_SOLVE via gaudi_envs mechanism and add it to env_variables.md

iboiko-habana (approved):

PR #1777, opened 2026-09-02
GLM-5.2-FP8 with DSA enablement

jkaniecki:
Enable GLM-5.2 with DSA (DeepSeek Sparse Attention) on HPU

What changed:

hpu_sparse_attn_indexer.py (new): per-request Q·K BF16 scoring + torch.topk to select top-2048 KV cache slots per decode step
hpu_attn.py: forward_mqa_sparse — gathers top-K latent KV, decompresses, runs masked MLA decode attention
oot_mla.py: dispatches to forward_mqa_sparse when use_sparse=True
deepseek_v2.py: BF16 indexer cache, HPU-safe Indexer.forward, SparseAttnIndexer dispatch
platform.py / hpu_model_runner.py: DSA routing + indexer cache layer index fix
Validated on 8× Gaudi3 (GLM-5.2-FP8, TP=8):

GSM8K 5-shot (1319 samples): 93.6% exact match
AIME 2026 - 100% - on pair with model card's target (99,2 % mean for multiple reruns)


vllm_gaudi/ops/hpu_sparse_attn_indexer.py
```diff
@@ -0,0 +1,86 @@
...
+    """HPU SparseAttnIndexer: per-request QK BF16 scoring + torch.topk."""
+    forward_context = get_forward_context()
+    attn_metadata = forward_context.attn_metadata
+    kv_cache = self.k_cache.kv_cache
+    if isinstance(kv_cache, tuple):
+        kv_cache = kv_cache[0]
+    block_size = attn_metadata.block_size
+    slot_mapping = attn_metadata.slot_mapping.flatten()
+
+    if kv_cache is None or kv_cache.numel() == 0:
```
pawel-olejniczak, line 27: On this branch I believe this early return is the only path this function ever takes. `get_kv_cache_spec` in `vllm_gaudi/v1/worker/hpu_model_runner.py:1717` handles only `MambaBase`, `Attention` and `MLAAttention`, and `DeepseekV32IndexerCache` is none of those - it is a plain `AttentionLayerBase` (upstream `vllm/model_executor/models/deepseek_v2.py:604`). So the indexer cache layer never gets a spec, never gets an allocation, and `self.k_cache.kv_cache` keeps the `torch.tensor([])` it is initialised with at `deepseek_v2.py:609`. `kv_cache.numel() == 0` is then true on every step: the `index_copy_` below never inserts K, and the returned indices are always `0..topk_tokens-1`.

The main-branch sibling (`#1760`) adds exactly the missing branch to `get_kv_cache_spec`, and this PR is otherwise byte-identical to it in all five shared files. The new `check_runner_kv_caches_multi_layer` no-op in `platform.py` only has an effect once the indexer cache is in `kv_caches`, which also points at that change being intended here.

Please cherry-pick the `elif isinstance(attn_module, AttentionLayerBase)` branch from `#1760`. One thing worth confirming when you do: it adds a second entry per DSA layer to `runner_kv_caches`, which is the case upstream's own TODO at `vllm/v1/worker/utils.py:529-541` describes as unanalyzed - have you checked that nothing on the HPU side indexes `runner_kv_caches` positionally?

vllm_gaudi/attention/oot_mla.py
```diff
@@ -150,11 +165,14 @@ def forward_impl(
         if is_prefill:
             output = self.impl.forward_mha(q, latent_vec_k, kv_cache, attn_metadata)
             return output
+        elif self.use_sparse and getattr(self, 'topk_indices_buffer', None) is not None:
```
pawel-olejniczak: `self.topk_indices_buffer` is never set on this object, so this branch cannot be taken. Upstream `MLAAttention.__init__` only forwards `topk_indices_buffer` into `extra_impl_args` for the impl (`vllm/model_executor/layers/attention/mla_attention.py:513-515`); it never assigns it to the module. The only assignment in this repo is line 327 of this file, which is on `HPUMultiHeadLatentAttentionWrapper` - a different object, since this class is its `self.mla_attn`. And `HPUMLAImpl.__init__` swallows the kwarg in `**kwargs` (`hpu_attn.py:192`) without storing it either.

So `use_sparse` is True, `getattr` returns None, and every DSA decode falls through to the dense `forward_mqa` on line 173, leaving `forward_mqa_sparse` unreachable.

Suggest storing the buffer on the impl in `HPUMLAImpl.__init__` and reading `self.impl.topk_indices_buffer` here. While you are there, `HPUMLAImpl` does not override `is_sparse`, so it inherits `ClassVar[bool] = False` from `vllm/v1/attention/backend.py:860` and advertises itself as dense; setting it keeps the impl and the module consistent. Note that fixing this gate is what makes the two issues I flagged on `hpu_sparse_attn_indexer.py:47` and `hpu_attn.py:406` live, so they need to land together rather than one at a time.

vllm_gaudi/ops/hpu_sparse_attn_indexer.py
```diff
@@ -0,0 +1,86 @@
...
+        return self.topk_indices_buffer
+
+    batch_size = q.shape[0]
+    block_list = attn_metadata.block_list
+
+    seq_lens = getattr(attn_metadata, "seq_lens_tensor", None)
+    if seq_lens is None:
+        context_lens = getattr(attn_metadata, "context_lens_tensor", None)
+        seq_lens = context_lens + 1 if context_lens is not None else None
+    if seq_lens is None:
```
pawel-olejniczak: Both fields this chain probes are always `None` on HPU decode, so this fallback is the live path and the QK scoring loop on lines 56-84 is unreachable. `HPUAttentionMetadataV1.make_decode_metadata` hardcodes `context_lens_tensor=None` (`vllm_gaudi/v1/attention/backends/hpu_attn.py:144`), and `_create_decode_input_data` builds `seq_lens_tensor` only when `self.num_mamba_like_layers > 0` (`hpu_model_runner.py:3291-3328`), which is 0 for GLM-5.2.

The fallback then hands every request in the batch the same first `topk_tokens` slots taken from the front of `block_list`, which is the padded concatenation over all requests, so for most rows those are other requests' slots. The same `getattr` chain in `forward_mqa_sparse` (`vllm_gaudi/attention/backends/hpu_attn.py:386-392`) falls back to `torch.full(..., topk)`, making `valid_topk == topk` for every row and the pad mask inert too.

Per-request lengths are derivable on device from `block_groups` and `block_usage`, both of which are populated for decode, and that also removes the `.item()` sync on line 60. Worth noting that the scoring loop's `block_offset` walk additionally assumes `block_list` is an unpadded per-request concatenation, which `block_groups` and the padding in `hpu_model_runner.py:2502` contradict.

pawel-olejniczak (commented):
Reviewed at 867254bc against VLLM_STABLE_COMMIT 2cf0a691.

The main thing to resolve before this lands: I do not think the sparse path can execute on this branch. There are three independent gates, any one of which routes every DSA decode back to dense MLA - see the inline comments on `oot_mla.py:168`, `hpu_sparse_attn_indexer.py:27` and `hpu_sparse_attn_indexer.py:47`. The first is that this backport is missing the `hpu_model_runner.py` change that the main-branch sibling carries; the other five files here are byte-identical to it.

That also affects how I read the validation numbers. Dense MLA is exact for sequences up to `index_topk` tokens, so GSM8K and AIME agreeing with the model card is consistent with the sparse path never running, and does not by itself show that it did. Something that separates the two paths would help, for example a log line or counter proving `forward_mqa_sparse` was entered, or an eval at a context length well past `index_topk`.

Two other asks:

- Performance numbers. DSA is a latency and throughput feature, and the description only reports accuracy, while the dense-fallback PR it supersedes reported median TTFT 1.82 s / TPOT 88.5 ms / 353.7 tok/s. Once the scoring path is actually reached it is a Python loop over the batch with a per-request `int(seq_lens[i].item())` host sync, inside `@torch.compiler.disable`. Could you post TPOT and output throughput at a long context next to that dense baseline?
- `pre-commit` will fail as is: the pinned yapf 0.43.0 rewrites 12 hunks across four of the five files. Pinned ruff 0.11.7 is clean.

vllm_gaudi/attention/backends/hpu_attn.py
```diff
@@ -353,19 +353,75 @@ def forward_mqa(  # type: ignore
...
+                    >= valid_topk.unsqueeze(1))
+        # Use finite minimum value to avoid NaN from softmax(-inf / -inf)
+        # when valid_topk == 0 (entire row masked)
+        attn = attn.masked_fill(
+            pad_mask.unsqueeze(1).unsqueeze(2), torch.finfo(attn.dtype).min)
+
+        attn = torch.softmax(attn * self.scale, dim=-1)
+        # Zero output for rows where entire sequence was masked (valid_topk == 0)
+        empty_mask = (valid_topk == 0).view(batch_size, 1, 1)
+        attn = attn * (~empty_mask).float()
```
pawel-olejniczak: `(~empty_mask).float()` is float32, so this multiply promotes `attn` from bfloat16 to float32, and the `torch.matmul(attn, v)` on the next line is then asked to mix a float32 `attn` with a bfloat16 `v`. `torch.matmul` does not type-promote, and there is no `autocast` anywhere in `vllm_gaudi` to paper over it. Use `.to(attn.dtype)` instead of `.float()`.

vllm_gaudi/models/deepseek_v2.py
```diff
@@ -87,24 +90,65 @@ def _hpu_deepseek_v2_model_forward(
...
+
 
-    return _orig_deepseek_v2_model_load_weights(self, _filtered(weights))
+def _hpu_sparse_indexer_forward_native(self, hidden_states, q_quant, k, weights):
+    from vllm_gaudi.ops.hpu_sparse_attn_indexer import forward_hpu
+    return forward_hpu(self, hidden_states, q_quant, k, weights)
 
 
-deepseek_v2.DeepseekV2Model.load_weights = _hpu_deepseek_v2_model_load_weights
+SparseAttnIndexer.forward_native = _hpu_sparse_indexer_forward_native
```
pawel-olejniczak, line 147: This PR drops `_hpu_deepseek_v2_model_load_weights`, which filtered `.indexers_proj.` out of the checkpoint. That filter was added for this exact model by a515ecd4 ("Enable GLM-5.2 (DSA MoE) on HPU via dense-MLA fallback").

At the pinned `VLLM_STABLE_COMMIT` (2cf0a691) upstream vLLM has no reference to `indexers_proj` at all, the `.indexer.` skip in `DeepseekV2Model.load_weights` (`vllm/model_executor/models/deepseek_v2.py:219-222`) does not match `indexers_proj`, and a name that matches nothing reaches `param = params_dict[name]` and raises `KeyError`. Nothing in this PR adds a module for those weights either.

Does the GLM-5.2-FP8 checkpoint you validated still ship `indexers_proj` weights, and if so, what consumes them now?

vllm_gaudi/ops/hpu_sparse_attn_indexer.py
```diff
@@ -0,0 +1,86 @@
...
+
+import math
+import torch
+from vllm.forward_context import get_forward_context
+
+
+def _fill_sequential(buf, n, topk_tokens, device):
+    """Fill topk_indices_buffer rows 0..n-1 with sequential slot indices."""
+    # Fill entire buffer width to avoid stale/out-of-range indices
+    buf[:n, :topk_tokens] = torch.arange(
```
pawel-olejniczak: Grouped nits:

- Formatting: the repo pins yapf 0.43.0 with `column_limit = 120` (`.pre-commit-config.yaml`, `pyproject.toml:42-43`), but the new code is wrapped at roughly 80 columns. Running the pinned yapf rewrites 12 hunks across `hpu_sparse_attn_indexer.py`, `models/deepseek_v2.py`, `attention/backends/hpu_attn.py` and `attention/oot_mla.py`, so `pre-commit` will fail. Pinned ruff 0.11.7 is clean.
- `_orig_forward_native` (`models/deepseek_v2.py:146`) is assigned and never used.
- `attention/oot_mla.py:177` has an orphaned `# during each graph execution` comment, left behind when the `NOTE(Xinyu)` line above it was deleted.
- `check_runner_kv_caches_multi_layer` (`platform.py:102`) is missing the `-> None` annotation that the four upstream platform overrides carry.
- None of the four commits carry `Signed-off-by`, which is why DCO is red.

iboiko-habana:
@jkaniecki do we need README and examples update?

vllm_gaudi/models/deepseek_v2.py
jkaniecki: I'm not sure what was the purpose of this filter before - I've checked the GLM-5.2-FP8 model and it doesn't contain any `indexers_proj`
```
grep -o 'indexers_proj' model.safetensors.index.json | wc -l
→ 0
```
It contains indexer projestion weight under: 
`model.layers.N.self_attn.indexer.weights_proj.weight`

vllm_gaudi/ops/hpu_sparse_attn_indexer.py
jkaniecki: runner_kv_caches is not consumed by any function / read by any other variable. 
kv_caches is also binded to model_runner.kv_Caches, which is consumed by defragmenter. I switched off the defrag for this model (and cont pa)  in features.py - we don't want it to manipulate indexer cache. model_runner.kv_caches is also used in hpu_worker.py:L531-550, but there only first item from kv cache is taken and it's not indexer cache.

vllm_gaudi/attention/oot_mla.py
jkaniecki: Fixed

vllm_gaudi/ops/hpu_sparse_attn_indexer.py
jkaniecki: Changed

vllm_gaudi/attention/backends/hpu_attn.py
jkaniecki: Fixed

vllm_gaudi/ops/hpu_sparse_attn_indexer.py
jkaniecki: Fixed

jkaniecki:
> Reviewed at [867254b](https://github.com/vllm-project/vllm-gaudi/commit/867254bc00b678df0491e840bea5b276d9855a17) against VLLM_STABLE_COMMIT 2cf0a691.
> 
> The main thing to resolve before this lands: I do not think the sparse path can execute on this branch. There are three independent gates, any one of which routes every DSA decode back to dense MLA - see the inline comments on `oot_mla.py:168`, `hpu_sparse_attn_indexer.py:27` and `hpu_sparse_attn_indexer.py:47`. The first is that this backport is missing the `hpu_model_runner.py` change that the main-branch sibling carries; the other five files here are byte-identical to it.
> 
> That also affects how I read the validation numbers. Dense MLA is exact for sequences up to `index_topk` tokens, so GSM8K and AIME agreeing with the model card is consistent with the sparse path never running, and does not by itself show that it did. Something that separates the two paths would help, for example a log line or counter proving `forward_mqa_sparse` was entered, or an eval at a context length well past `index_topk`.
> 
> Two other asks:
> 
> * Performance numbers. DSA is a latency and throughput feature, and the description only reports accuracy, while the dense-fallback PR it supersedes reported median TTFT 1.82 s / TPOT 88.5 ms / 353.7 tok/s. Once the scoring path is actually reached it is a Python loop over the batch with a per-request `int(seq_lens[i].item())` host sync, inside `@torch.compiler.disable`. Could you post TPOT and output throughput at a long context next to that dense baseline?
> * `pre-commit` will fail as is: the pinned yapf 0.43.0 rewrites 12 hunks across four of the five files. Pinned ruff 0.11.7 is clean.

@pawel-olejniczak Thanks for the complex review, I've changed the code to enable DSA properly, confirmed it is now launching. About testing - for sure GSM8K was only a sanity check, not showing real DSA capabilities. AIME 2026 is a bit different - some samples overcome this 2k topk threshold and switch on the DSA, so it's really working. I've also ran tests up to 128k context to see long-context capabilities using "needle" type task - if was successful in all runs. 

About perf numbers - it's only a functional enablement for now, so not sharing perf results, I haven't focus on them. The MLA fallback PR may have some perf numbers (even better than DSA version due to switching t.compile off) but in fact it enables different model - GLM5.2 was trained and should be executed with DSA. Without it, we cannot be sure about it's accuracy for samples larger than top_k. Dense MLA version can also show good acc at multiple sample ranges, but it cannot be said that it's the same acc that GLM5.2 should have then.

vllm_gaudi/attention/backends/hpu_attn.py
```diff
@@ -260,6 +260,10 @@ def __init__(
                                                  f"heads in the layer. Sinks shape: {sinks.shape}, "
                                                  f"num_heads: {num_heads}.")
 
+
+        self.topk_indices_buffer = kwargs.get('topk_indices_buffer')
```
pawel-olejniczak, line 264: Small leftover from the formatting pass: there are two blank lines after the `sinks` assert, and the pinned yapf 0.43.0 removes one, so `pre-commit` still fails on this file. The other four files are clean now.

    @@ -259,7 +259,6 @@
                                                      f"num_heads: {num_heads}.")
    -
     
             self.topk_indices_buffer = kwargs.get('topk_indices_buffer')

Also `_fill_invalid(buf, n, device)` in `hpu_sparse_attn_indexer.py:7` no longer uses its `device` argument after the switch to `fill_(-1)`.

pawel-olejniczak (approved):
Re-reviewed at 09f10b54. The three gates are closed as far as I can verify by reading: the `get_kv_cache_spec` branch is in, the buffer now lives on the impl where `forward_impl` reads it, and deriving per-request slots from `block_groups` plus `block_usage` is a better fix than the `seq_lens` plumbing I suggested. I also confirmed defrag follows `use_contiguous_pa`, so disabling contiguous PA for `glm_moe_dsa` covers both. Thanks for the checkpoint check on `indexers_proj`.

Approving. Only the yapf nit below is left from my side, and it will keep `pre-commit` red until it is fixed.

PR #1795, opened 2026-09-14
[Bugfix] Support MLA variants with qk_rope_head_dim == 0

ArcheyChen:
## Purpose

Some MLA checkpoints carry no RoPE component. GLM-5.3-Flash is one — `qk_rope_head_dim=0` with `qk_nope_head_dim=256`, so `qk_head_dim == qk_nope_head_dim`.

`HPUAttentionImpl.forward_mha` splits `latent_vec_k` into `(k_c_normed, k_pe)` and reshapes the RoPE half:

```python
k_c_normed, k_pe = latent_vec_k.split([self.kv_lora_rank, self.qk_rope_head_dim], dim=-1)
k_pe = k_pe.view(-1, 1, self.qk_rope_head_dim)
```

With a zero-sized RoPE part the split is fine and yields an empty tensor, but the reshape is not:

```
RuntimeError: cannot reshape tensor of 0 elements into shape [-1, 1, 0]
because the unspecified dimension size -1 can be any value and is ambiguous
```

Worth spelling out which operations are and are not safe on a zero-sized dim, since only one of them actually breaks:

| op | zero-sized dim |
|---|---|
| `split([512, 0])` | fine |
| `cat([k_nope, empty])` | fine |
| `view(-1, 1, 0)` | **ambiguous, raises** |

So the fix only guards the reshape and the concat that consumes it. With no RoPE part, `k` is simply `k_nope`.

## Test Plan

`tests/unit_tests/test_mla_no_rope.py`, CPU-only, no accelerator needed:

- `test_k_assembly_handles_zero_rope_dim` — parametrized over `qk_rope_head_dim` in `{0, 64}`, checks `k` comes out as `[tokens, heads, qk_nope + qk_rope]` for both
- `test_unguarded_reshape_is_what_fails` — pins down that `split` and `cat` are fine and the ambiguous `view` is the actual failure, so the distinction does not get lost in a future refactor

```bash
pytest tests/unit_tests/test_mla_no_rope.py
```

## Test Result

```
3 passed
```

End-to-end: this was one of the changes needed to bring GLM-5.3-Flash up on Intel Gaudi 2 (TP=4). Without it the server crashes during warmup; with it (plus related fixes) the model serves and produces correct output.

`ruff check` passes on both files.

---

*AI disclosure: this change was developed with assistance from Claude (Claude Code). I reviewed every changed line, reproduced the crash and validated the fix on real Gaudi 2 hardware, and ran the tests myself.*


vllm_gaudi/attention/backends/hpu_attn.py
```diff
@@ -278,14 +278,22 @@ def forward_mha(  # type: ignore
         # =========================== #
 
         k_c_normed, k_pe = latent_vec_k.split([self.kv_lora_rank, self.qk_rope_head_dim], dim=-1)
-        k_pe = k_pe.view(-1, 1, self.qk_rope_head_dim)
+        # MLA variants without a RoPE component (qk_rope_head_dim == 0) leave
+        # k_pe empty; view(-1, 1, 0) is ambiguous because a zero in the shape
+        # makes the element count 0 and -1 unsolvable.
+        if self.qk_rope_head_dim > 0:
+            k_pe = k_pe.view(-1, 1, self.qk_rope_head_dim)
```
pawel-olejniczak: `k_pe` here is always 2-D (`latent_vec_k` is forced to `[-1, kv_lora_rank + qk_rope_head_dim]` at `vllm_gaudi/attention/oot_mla.py:158`, and the prefix-cache branch above keeps it 2-D), so `view(-1, 1, self.qk_rope_head_dim)` is exactly `unsqueeze(1)`. Writing it that way is unambiguous for `qk_rope_head_dim == 0` as well, which removes the need for this first guard: with no RoPE part the resulting empty `[tokens, 1, 0]` tensor is simply never consumed by the branch below. I checked both forms on CPU torch 2.11: for `qk_rope_head_dim=64` the assembled `k` is bit-identical, and for `0` the `unsqueeze` form returns `[tokens, 1, 0]` instead of raising.

```python
k_pe = k_pe.unsqueeze(1)
```

Separately: with the explicit shape in place the second branch also looks droppable, which keeps `k` contiguous (`k = k_nope` is a strided slice of `kv_nope`, unlike every `qk_rope_head_dim > 0` path, which has always fed a freshly concatenated tensor into `ops.prompt_attention`). I measured that rather than leaving it as a question: on a Gaudi 3 card in eager mode (`PT_HPU_LAZY_MODE=0`, bf16) dropping both branches gives a result bit-identical to this patch (`max_abs_diff = 0.0`) through both `fsdpa_impl` and `naive_impl`, for `qk_rope_head_dim` 64 and 0, and it survives `torch.compile(backend="hpu_backend", dynamic=False)` with an identical second-call replay.

One gap I could not close: lazy mode. The stack I have refuses `PT_HPU_LAZY_MODE=1` on upstream torch 2.11, so a zero-width `cat` captured by `wrap_in_hpu_graph` is untested, and you are on Gaudi 2 rather than Gaudi 3. If that is what your run uses and it is why you wrote the branch, it is worth a one-line comment saying so.

tests/unit_tests/test_mla_no_rope.py
```diff
@@ -0,0 +1,78 @@
...
+import torch
+
+KV_LORA_RANK = 512
+QK_NOPE_HEAD_DIM = 256
+V_HEAD_DIM = 256
+NUM_HEADS = 16
+TOKENS = 4
+
+
+def _k_from_latent(latent_vec_k, qk_rope_head_dim, guarded):
```
pawel-olejniczak: Trivia: the `guarded=False` path of this helper (lines 39-40) is never reached. Both tests in the file call `_k_from_latent(..., guarded=True)` or do not call it at all, so the parameter and its `else` branch are dead. Dropping the parameter would make the helper say what it actually tests.

pawel-olejniczak (commented):
Reviewed at `c379e1b`. The fix is correct and, as far as I can tell, complete: I enumerated every use of `qk_rope_head_dim` in the plugin and `hpu_attn.py:285` was the only ambiguous `-1` reshape. `oot_mla.py:157` already passes explicit sizes so `view(tokens, 0)` is well defined there, and the decode path is safe by construction (`torch.cat([q_nope, q_pe])` with an empty `q_pe`, then `value = key[..., :kv_lora_rank]` in `extension/ops.py:185`). `qk_head_dim == qk_nope_head_dim` when the RoPE dim is zero also holds upstream (`qk_head_dim = qk_nope_head_dim + qk_rope_head_dim`), so the comment on the new `else` branch is accurate.

Two notes below, neither blocking. Nothing to say about style: ruff 0.11.7 and yapf 0.43.0 at the repo's pinned versions are both clean on the changed files, and the `# noqa: SIM108` is load-bearing (SIM108 does fire without it).

Two small things that do not fit an inline anchor:

- The PR description attributes `forward_mha` to `HPUAttentionImpl`; the method is on `HPUMLAImpl` (`hpu_attn.py:169`, `:263`). `HPUAttentionImpl` is a different class further down the file.
- The pre-merge `gate` and `execute_pre_merge` runs on this SHA are both `cancelled`, so the HPU unit-test job has not actually run this test yet.

ArcheyChen:
Thanks — that is a much better patch than mine. Taken in full: `k_pe.unsqueeze(1)`
and both branches gone, so the production change is one line plus its comment.
New SHA is `696f4af`.

**Your question: no, I never hit a zero-width `cat`.** The bring-up ran
`--enforce-eager` (TP=4, Gaudi 2), so the concat branch was defensive, not
empirical — I guarded it because it was adjacent to the reshape, not because I
saw it fail. My own notes from that day even record `cat` as safe on a
zero-sized dim, so the branch contradicted what I had already measured. There is
no comment to add, because there is nothing to justify.

Your contiguity argument is what settles it, and it points the other way from
the patch I wrote. Reproduced on CPU, torch 2.13, assembling `k` the three ways:

```
rope=64 view       k=(4, 16, 320) contiguous=True
rope=64 unsqueeze  k=(4, 16, 320) contiguous=True
rope=64 branch     k=(4, 16, 320) contiguous=True
rope= 0 view       RuntimeError: cannot reshape tensor of 0 elements into shape [-1, 1, 0] ...
rope= 0 unsqueeze  k=(4, 16, 256) contiguous=True
rope= 0 branch     k=(4, 16, 256) contiguous=False
rope=64 view vs unsqueeze bit-identical: True
rope= 0 unsqueeze vs branch bit-identical: True
```

Same values, and `k = k_nope` was handing `ops.prompt_attention` a strided slice
of `kv_nope` — the one thing on that path no `qk_rope_head_dim > 0` run has ever
done.

**Lazy mode: I cannot close that gap either.** The Gaudi 2 was rented for the
GLM-5.3-Flash bring-up and I no longer have it, so I have no stack to capture a
zero-width `cat` under `wrap_in_hpu_graph`. I would rather leave it recorded as
untested than imply otherwise. What I can say is that it cannot regress anything
that works today: at `qk_rope_head_dim == 0` this path raises at the reshape
before it ever reaches the concat, so the only users of the concat form are the
`> 0` paths that already use it.

**Test.** The `guarded` parameter is gone — with no branch left in production
there is nothing for it to select, and the helper now mirrors the production
form exactly. It also picked up the contiguity assertion, since that is a
property of the fix worth pinning.

On the earlier bot comment that the test does not drive `HPUMLAImpl.forward_mha`
itself: that is a fair criticism and I am not able to act on it. Importing
`vllm_gaudi.attention.backends.hpu_attn` pulls in `vllm_gaudi.extension.*` and
therefore `habana_frameworks`, so the test cannot run off-HPU, and without the
card I cannot execute it anywhere — I would be adding a test I have never run.
If you would rather have it, say so and I will write it against `HPUMLAImpl`
with the cache and projection mocked, with the understanding that your HPU
unit-test job is the first place it will actually execute.

Fixed the `HPUAttentionImpl` / `HPUMLAImpl` slip in the description; thank you
for catching it. The pre-merge runs were indeed cancelled on the old SHA, so
this test has not run in CI yet.

pawel-olejniczak (approved):

PR #1796, opened 2026-09-14
[Bugfix] Translate global expert ids to local ids under expert parallelism

ArcheyChen:
## Purpose

`torch.ops.hpu.mixture_of_experts` indexes its local weight list directly by the entry it finds in `expert_routing_table`. The table must therefore hold **local** expert ids (`0..local_num_experts-1`, with `-1` for tokens owned by another rank), and `experts_min`/`experts_max` must be the local range.

`HPUFp8MoEMethod` passed **global** ids together with a **global** window:

```python
ep_shift = layer.moe_config.ep_rank * num_experts
experts_min, experts_max = ep_shift, num_experts + ep_shift - 1
```

and never consumed `layer.expert_map` — the global-to-local table `ExpertMapManager` already builds. (`grep -r expert_map vllm_gaudi/` finds no consumers.)

### Why it went unnoticed

For `ep_rank == 0` the two conventions coincide, so rank 0 is correct and single-rank or non-EP runs look fine. Only ranks 1..N-1 are affected:

```
8 global experts, EP=4, 2 local per rank, routing [2, 3, 0, 5, 2, 7]

rank0 (owns global [0])      -> correct
rank1 (owns global [2,3,2])  -> ids fall outside its local list, all three dropped
rank2 (owns global [5])      -> dropped
rank3 (owns global [7])      -> dropped
```

Nothing raises. With GLM-5.3-Flash (288 experts, top-8, 42 MoE layers) on 4-way EP the lost contributions average out across layers and generation collapses to a repeated token.

I verified the kernel's convention directly on a card before writing the fix: driving `torch.ops.hpu.mixture_of_experts` with 8 global / 2 local experts, the current calling convention produced `0.0000` for tokens the rank owned and non-zero values for tokens it did not; switching to local ids plus the `-1` sentinel made ranks 1/2/3 match the reference bit-for-bit (~1e-4, pure bf16 noise).

## Test Plan

`tests/unit_tests/test_ep_expert_map.py`, CPU-only — it models the kernel's selection rule in plain PyTorch so the convention is pinned down without a card or a distributed run:

- `test_local_ids_route_to_the_owned_experts` — parametrized over all 4 ranks
- `test_expert_map_marks_remote_experts_with_minus_one` — the sentinel is what keeps non-local tokens out
- `test_untranslated_global_ids_route_wrongly` — regression guard, ranks 1..3 only

```bash
pytest tests/unit_tests/test_ep_expert_map.py
```

## Test Result

```
8 passed
```

End-to-end on Intel Gaudi 2, GLM-5.3-Flash, TP=4 + EP=4, greedy decoding, prompt `中国的首都是` ("The capital of China is"):

| | output | s/token |
|---|---|---|
| before | `首首首首首首首首` | 9.1 |
| after | `北京，北京是中国的政治、文化` | 8.6 |
| TP=4, no EP (reference) | `北京，北京是中国的政治、文化` | 26 |

Output with EP now matches the non-EP path exactly, and the non-EP path is unchanged (`expert_map` is `None` when EP is off, so the translation is a no-op there — verified by re-running the non-EP config after the change).

### Side benefit

The kernel's cost scales with the number of experts handed to it, not with top-k. Measured per MoE layer on Gaudi 2:

| experts per rank | time |
|---|---|
| 288 (no EP) | 163-175 ms |
| 288 split into 5x64 slices | 179-182 ms (slower) |
| 72 (EP=4) | **47 ms** |

Across 42 MoE layers that is ~6.8 s/token versus ~2.0 s/token, which is where the 26 -> 8.6 s/token end-to-end improvement comes from.

`ruff check` passes on both files.

---

*AI disclosure: this change was developed with assistance from Claude (Claude Code). I reviewed every changed line, verified the kernel's calling convention on real Gaudi 2 hardware, validated the fix end-to-end including a non-EP regression run, and ran the tests myself.*


vllm_gaudi/ops/hpu_fp8.py
```diff
@@ -283,9 +283,13 @@ def create_weights(self, *args, **kwargs) -> None:
 
     def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
         num_experts = layer.local_num_experts
-        ep_shift = layer.moe_config.ep_rank * num_experts
 
-        experts_min, experts_max = ep_shift, num_experts + ep_shift - 1
+        # The Habana MoE kernel indexes its expert slots by LOCAL id, so the
```
pawel-olejniczak, line 287: This comment states the kernel indexes its expert slots by local id, but the
repo's own usage says the kernel subtracts `experts_min` itself. In the sliced
branch at `vllm_gaudi/extension/ops.py:1296-1310` the op is handed
`w13_list[i*n:(i+1)*n]` (n tensors) together with `experts_min + i*n .. +n-1`;
that can only select `slice[j]` for routing id `experts_min + i*n + j` if the
kernel maps `slot = id - experts_min`. `vllm_gaudi/ops/hpu_fused_moe.py:114`
documents `experts_min` the same way ("first global expert id owned by this (EP)
rank"), and both pure-PyTorch stand-ins for the op implement `local = global -
experts_min` (`hpu_fused_moe.py:116`, `hpu_moe_combine.py:76`).

If that is right, the old and new calling conventions are equivalent for the
stock op, ranks 1..N-1 were not dropping their own experts, and
`tests/unit_tests/test_ep_expert_map.py:41` encodes the disputed premise as
ground truth (it checks `0 <= rid < LOCAL_EXPERTS` instead of
`0 <= rid - experts_min < LOCAL_EXPERTS`), which is why the regression guard
passes. What did the standalone repro pass as `experts_min`/`experts_max`, and
how many weight tensors were in `w12`/`w3`?

There is still a real defect underneath, and `expert_map` is the right fix for
it: `ep_shift = ep_rank * local_num_experts` only equals this rank's global base
when `global_num_experts % ep_size == 0` and placement is linear. With
`global_num_experts = 290, ep_size = 4`, `determine_expert_map` gives rank 2
`start_idx = 146` while `ep_shift = 144`, so every owned token reads the weights
two slots down and globals 216-217 fall outside the window; with
`--expert-placement-strategy round_robin` the local ids are not contiguous at
all and the arithmetic is meaningless. Could you re-frame the description and
this comment around that (the window is not wrong, `ep_shift` is), and confirm
the expert count and placement strategy of the GLM run so the measurement lines
up with the stated cause?

tests/unit_tests/test_ep_expert_map.py
```diff
@@ -0,0 +1,82 @@
...
+
+GLOBAL_EXPERTS = 8
+EP_SIZE = 4
+LOCAL_EXPERTS = GLOBAL_EXPERTS // EP_SIZE
+ROUTING = [2, 3, 0, 5, 2, 7]
+
+
+def build_expert_map(ep_rank: int) -> torch.Tensor:
+    """What ExpertMapManager builds: global -> local, -1 when not local."""
+    m = torch.full((GLOBAL_EXPERTS,), -1, dtype=torch.long)
```
pawel-olejniczak, line 30: This file is not clean under the repo's pinned formatter. `yapf` 0.43.0 with
`pyproject.toml`'s `column_limit = 120` rewrites 5 lines here - `(GLOBAL_EXPERTS,)`
to `(GLOBAL_EXPERTS, )` on this line, and the slice spacing on lines 32, 67 and
69 (`m[lo : lo + LOCAL_EXPERTS]` to `m[lo:lo + LOCAL_EXPERTS]`). The PR body
mentions `ruff check`, which does pass, but `ruff-format` is scoped to
`.buildkite|benchmarks|examples` in `.pre-commit-config.yaml`, so `yapf` is the
formatter for this path and the `pre-commit` job gates
`pre_merge_hpu_test_build`. Please run `pre-commit run --all-files`.

pawel-olejniczak (commented):
Thanks for the detailed writeup and the hardware validation - the end-to-end
before/after is exactly the right evidence to bring. I have one question about
the stated root cause and one CI blocker; I have not repeated the three points
Copilot already made, but I agree with the gather-path and int32 ones and think
both need to land.

One note rather than a separate comment: if `expert_map` becomes the source of
truth here, the same `ep_shift = ep_rank * local_num_experts` computation also
lives at `vllm_gaudi/ops/hpu_fused_moe.py:412`, `vllm_gaudi/ops/hpu_mxfp4.py:70`
and `vllm_gaudi/ops/hpu_compressed_tensors.py:444` and `:963`. Worth deciding
whether this PR scopes to FP8 only (fine by me, just say so) or whether the
translation belongs in one shared helper.

PR #1797, opened 2026-09-14
[Gemma4] Add bidirectional (prefix-LM) image mask on HPU prefill

jiminha:
**What & why**
Gemma4 26B/31B/12B set use_bidirectional_attention == "vision". On these, vLLM applies a prefix-LM image mask: tokens within a single image's span attend bidirectionally (only on sliding-attention layers, clamped to the window), while generated text stays causal. The HPU plugin had none of this — every prefill built a pure causal triangle, so image patches were blinded to ~half the patches of their own image, degrading multimodal output. This brings HPU to parity with CUDA.

**Changes**
Fresh-prefill mask — bake allowed = (causal ∧ sw_left) ∨ (mm_block ∧ sw_left) into the FusedSDPA sliding-window bias. A per-image segment-id map keeps cross-image attention forbidden; audio spans stay causal.
Chunked prefill — apply the same term in the block_list branch (chunks 2..N and prefix-cache hits), and force disable_chunked_mm_input for Gemma4 (as platforms/cuda.py does) so an image span is never split across a chunk boundary.
Gated on model_config.is_mm_prefix_lm, scoped to Gemma4. Non-FusedSDPA path stays causal (documented).

**Validation (HPU vs B200)**
check	result
Mask unit test vs independent reference	19/19 pass (incl. chunk-2 ≡ single-shot)
20-image describe, 26B/31B	20/20, matches GPU exactly
Single-shot regression	byte-identical to pre-change
Perf	26B −2.5%, 31B −1.1% (mostly the span-atomic scheduler constraint; CUDA pays the same)


vllm_gaudi/v1/worker/hpu_model_runner.py
```diff
@@ -7600,13 +7684,23 @@ def _set_attn_bias_for_sliding_window(self, attn_metadata: HPUAttentionMetadataV
...
+        elif image_seg_ids is None:
             # CAUSAL MASK without removing padding (CAUSAL+sliding window)
             # removing padding cause accuracy issue for images input
             tensor = torch.full((batch_size, 1, seq_len, seq_len), device=device, dtype=dtype, fill_value=1)
             mask = torch.tril(tensor, diagonal=shift)
             mask = torch.triu(mask, diagonal=shift - window_size + 1)
             attn_bias = torch.log(mask)
+        else:
+            # Gemma4 bidirectional image mask on a fresh prefill.
+            # Full-attn layers stay pure causal.
```
iboiko-habana: Full-attn layers stay pure causal. contradicts upstream. In compute_kv_seq_mask, the if USE_MM_PREFIX: block is top-level and ends in seq_mask |= mm_mask; only the clamp is gated on MM_PREFIX_CLAMP_SW and SLIDING_WINDOW > 0. The docstring states the contract as (causal AND window) OR mm_prefix, and "Inert for full-attention layers" refers to the clamp, not the OR.

Gemma4 is interleaved, so its full-attn layers read attn_bias from _set_attn_bias, which this diff doesn't touch. Net effect: forward same-image attention works on sliding layers and is -inf on full-attn layers. Runs clean, attends wrong.

iboiko-habana, line 7700: hpu_model_runner.py:7552	This builder feeds attn_bias, which Gemma4's full-attention layers read (hpu_attn.py only overwrites with window_attn_bias when the layer has self.sliding_window). It is left pure causal, so forward same-image attention works on sliding layers and is -inf on full-attn layers — a divergence from CUDA. Upstream compute_kv_seq_mask runs seq_mask \|= mm_mask at top level; only the clamp is gated on SLIDING_WINDOW > 0. Fix = read image_seg_ids here, add and image_seg_ids is None to the fusedsdpa escape on this line (otherwise a fresh prefill returns attn_bias=None and gets built-in causal), then OR in the unclamped block. Polarity is inverted vs the sliding builder: mask here means blocked, so & ~block.



See [_set_attn_bias](vscode-file://vscode-app/c:/Users/iboiko/AppData/Local/Programs/Microsoft%20VS%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) — this comment states the gap. Upstream applies the block to full-attn layers too, unclamped; "inert for full-attention layers" in [compute_kv_seq_mask](vscode-file://vscode-app/c:/Users/iboiko/AppData/Local/Programs/Microsoft%20VS%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) refers to the sliding-window clamp, not the OR.

iboiko-habana (changes requested):

vllm_gaudi/v1/worker/hpu_model_runner.py
jiminha: @iboiko-habana  In CUDA, yes as you mentioned inside compute_kv_seq_mask, the seq_mask |= mm_mask OR isn't gated on SLIDING_WINDOW; only the clamp is. However the real gate is one level up, so full-attn layers don't actually get the image block on CUDA either, but stay causal on gemma4 model.  
      => Please check gemma4_mm.py : _clear_mm_prefix_for_full_attn_layers() 
So USE_MM_PREFIX is never TRUE in full attention layer in CUDA. 

HPU split matches that exactly. 
full-attn → causal, via _set_attn_bias
sliding → causal + window + clamped image block, via _set_attn_bias_for_sliding_window

iboiko-habana (approved):

